PR #1, opened 2017-11-13
Should use admin to run acceptance test

yongfengdu:
And remove the || true, otherwise, the FAIL will be ignored


kiwik:
Looks good to me, thank you. @yongfengdu

liu-sheng:
LGTM +1

PR #3, opened 2017-11-28
Fix the pike/mitaka acc test issue

yongfengdu:
Use admin to run tests
Remove the |true debug setting
Adjust the test cases


kiwik:
LGTM, thank you.

PR #5, opened 2017-11-30
Use init-test as all jobs parents

kiwik:
Using init-test jobs to ensure workspace exist in pre,
and copy log files on node in post.


liu-sheng:
looks good to me +1

PR #7, opened 2017-12-04
Set OS_IDENTITY_API_VERSION=3 in Mitaka jobs

kiwik:
The default value of OS_IDENTITY_API_VERSION is 2.0 in Mitaka devstack,
but V2.0 API of keystone will be deprecated in the future, test against
V3 API of keystone in jobs to keep it ok.


liu-sheng:
LGTM +1

PR #9, opened 2017-12-04
Set OS_AUTH_URL point to v3 URL in Mitaka jobs

kiwik:
The default value of OS_AUTH_URL looks like http://localhost:5000/v2.0,
set it point to v3 URL in Mitaka jobs http://localhost:5000/v3


liu-sheng:
LGTM +1

PR #10, opened 2017-12-05
Add OS_DOMAIN_NAME option for Mitaka jobs

kiwik:
We use Keystone V3 API in all the jobs, so OS_DOMAIN_NAME is back


liu-sheng:
LGTM +1

PR #11, opened 2017-12-07
Add Gophercloud Acc test with Telefonica cloud

liu-sheng:


liu-sheng:
quickly merging for testing

PR #17, opened 2018-01-23
Clean the useless files collection tasks

liu-sheng:
We don't need to collect tox, testr related logs.


liu-sheng:
have tested this, please see: https://github.com/liu-sheng/gophercloud/pull/9

kiwik:
Thank you, looks good.

PR #20, opened 2018-01-30
Refactor devstack local.conf and installation roles

animationzl:
- refactor devstack local.conf to populate config according to the openstack version and request services
- refactor devstack installation to keep simple and consistent


roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -1,11 +1,44 @@
...
+      # in the future, test against V3 API in stead.
+      IDENTITY_API_VERSION=3
+      OS_USER_DOMAIN_ID=default
+      OS_PROJECT_DOMAIN_ID=default
+      EOF
+    executable: /bin/bash
+    chdir: '{{ ansible_user_dir }}/workspace'
+  environment: '{{ zuul | zuul_legacy_vars }}'
+  when:
+    - os_branch == 'mitaka-eol'
```
kiwik: os_branch should be 'stable/mitaka'

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -1,11 +1,44 @@
...
-- name: create devstack local conf with {{ os_branch }} for {{ sdk }}
+# Populate local conf header
+- name: create devstack local conf header
   shell:
     cmd: |
       set -e
       set -x
       cat << EOF >> /tmp/dg-local.conf
       [[local|localrc]]
+      EOF
```
kiwik: NETWORK_GATEWAY=10.0.0.1 should be here, make all the jobs to use the same network configuration.

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -30,69 +63,34 @@
...
-
-# terraform-provider-openstack
-- name: create devstack local conf with {{ os_branch }} for {{ sdk }}
-  shell:
-    cmd: |
-      set -e
-      set -x
-      cat << EOF >> /tmp/dg-local.conf
-      [[local|localrc]]
       NETWORK_GATEWAY=10.0.0.1
```
kiwik: move NETWORK_GATEWAY=10.0.0.1 to common task section.

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -30,69 +63,34 @@
     chdir: '{{ ansible_user_dir }}/workspace'
   environment: '{{ zuul | zuul_legacy_vars }}'
   when:
-    - sdk == 'gophercloud'
-    - os_branch in stable_branches
+    - '"manila" in enable_services'
```
kiwik: Expression statement is wrapped in string, seems it always be true?

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -30,69 +63,34 @@
...
       enable_service designate,designate-central,designate-api,designate-pool-manager,designate-zone-manager,designate-mdns
       enable_plugin designate git://git.openstack.org/openstack/designate
       EOF
     executable: /bin/bash
     chdir: '{{ ansible_user_dir }}/workspace'
   environment: '{{ zuul | zuul_legacy_vars }}'
   when:
-    - sdk == 'terraform-provider-openstack'
-    - os_branch in stable_branches
+    - '"designate" in enable_services'
```
kiwik: ditto

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -30,69 +63,34 @@
...
-      export OS_PROJECT_DOMAIN_ID=default
+      enable_plugin openstack-cloud-controller-manager https://github.com/dims/openstack-cloud-controller-manager
       EOF
     executable: /bin/bash
     chdir: '{{ ansible_user_dir }}/workspace'
   environment: '{{ zuul | zuul_legacy_vars }}'
   when:
-    - sdk == 'terraform-provider-openstack'
-    - os_branch in eol_branches
+    - '"openstack-cloud-controller-manager" in enable_services'
```
kiwik: ditto

roles/create-devstack-local-conf/defaults/main.yaml
```diff
@@ -0,0 +1,4 @@
+---
+os_branch: 'master'
+enable_services:
+  - 'default'
```
kiwik: I prefer to show the default service list at here, that's easier to understand, like:

```yaml
enable_services:
  - 'nova'
  - 'neutron'
  - 'cinder'
  - 'keystone'
  - 'glance'
  - 'swift'

liu-sheng:
please review PR: https://github.com/theopenlab/openlab-zuul-jobs/pull/21

roles/create-devstack-local-conf/tasks/main.yml
animationzl: already fixed in the new RP

roles/create-devstack-local-conf/tasks/main.yml
animationzl: thanks for your advise, already fixed in the new RP

roles/create-devstack-local-conf/tasks/main.yml
animationzl: Nope, this is the string in variable list syntax required by ansible, see the reference here: https://stackoverflow.com/questions/36496911/run-an-ansible-task-only-when-the-variable-contains-a-specific-string

roles/create-devstack-local-conf/defaults/main.yaml
animationzl: valuable advice, thanks

PR #21, opened 2018-01-30
Refactor devstack local.conf and installation roles 

liu-sheng:
- refactor devstack local.conf to populate config according to the openstack version and request services
- refactor devstack installation to keep simple and consistent

### DIFF
--- a/roles/create-devstack-local-conf/tasks/main.yml
+++ b/roles/create-devstack-local-conf/tasks/main.yml
@@ -1,5 +1,5 @@
 # Populate local conf with specific branch
-- name: create devstack local conf with {{ os_branch }} compatible
+- name: create devstack local conf with '{{ os_branch }}' compatible
   shell:
     cmd: |
       set -e
@@ -20,7 +20,7 @@
 
 
 # Populate local conf with specific service
-- name: create devstack local conf with {{ enable_services }} enabled
+- name: create devstack local conf with '{{ enable_services }}' enabled
   shell:
     cmd: |
       set -e
@@ -53,7 +53,7 @@
   when:
     - '"manila" in enable_services'
 
-- name: create devstack local conf with {{ enable_services }} enabled
+- name: create devstack local conf with '{{ enable_services }}' enabled
   shell:
     cmd: |
       set -e
--- a/roles/create-devstack-local-conf/tasks/main.yml
+++ b/roles/create-devstack-local-conf/tasks/main.yml
@@ -51,7 +51,7 @@
     chdir: '{{ ansible_user_dir }}/workspace'
   environment: '{{ zuul | zuul_legacy_vars }}'
   when:
-    - 'manila' in enable_services
+    - '"manila" in enable_services'
 
 - name: create devstack local conf with {{ enable_services }} enabled
   shell:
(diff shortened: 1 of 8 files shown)


zuul.d/jobs.yaml
```diff
@@ -157,8 +138,6 @@
     description: |
```
kiwik: this designate job use the specific run.yaml, don't use its parents' script, so I think we can set parent as "golang-test".

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -50,49 +30,53 @@
...
+      enable_service designate,designate-central,designate-api,designate-pool-manager,designate-zone-manager,designate-mdns
+      enable_plugin designate git://git.openstack.org/openstack/designate
       EOF
     executable: /bin/bash
     chdir: '{{ ansible_user_dir }}/workspace'
   environment: '{{ zuul | zuul_legacy_vars }}'
   when:
-    - sdk == 'terraform-provider-openstack'
-    - os_branch in eol_branches
+    - '"designate" in enable_services'
```
kiwik: see my comments in PR #20

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -50,49 +30,53 @@
...
+      enable_plugin heat git://git.openstack.org/openstack/heat
+      enable_plugin manila git://git.openstack.org/openstack/manila
       EOF
     executable: /bin/bash
     chdir: '{{ ansible_user_dir }}/workspace'
   environment: '{{ zuul | zuul_legacy_vars }}'
   when:
-    - sdk == 'terraform-provider-openstack'
-    - os_branch in stable_branches
+    - '"manila" in enable_services'
```
kiwik: see my comments in PR #20

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -50,49 +30,53 @@
     chdir: '{{ ansible_user_dir }}/workspace'
   environment: '{{ zuul | zuul_legacy_vars }}'
   when:
-    - sdk == 'gophercloud'
-    - os_branch in eol_branches
+    - os_branch == 'mitaka-eol'
```
kiwik: see my comments in PR #20

zuul.d/jobs.yaml
animationzl: both 'golang-test' and 'terraform-provider-openstack-acceptance-test' as parent job will have no impact on the designate job execution however, I agree with you as choosing the 'golang-test' as parent job is more  logical.

roles/create-devstack-local-conf/tasks/main.yml
animationzl: Nope, this is the string in variable list syntax required by ansible, see the reference here: https://stackoverflow.com/questions/36496911/run-an-ansible-task-only-when-the-variable-contains-a-specific-string

roles/create-devstack-local-conf/tasks/main.yml
animationzl: ditto

PR #22, opened 2018-01-31
Fine tune task names

animationzl:
Fine tune task names to avoid confusion


liu-sheng:
looks good to me. thanks.

Issue #49, opened 2018-02-08
Enable Octavia in LBaaS test of terraform-openstack-provider


kiwik:
Set environment variable OS_USE_OCTAVIA=true in LBaaS job to make terraform knowing the deployment is LBaaS + Octavia

PR #54, opened 2018-02-08
Add enable service and plugin for LBaaS v2

kiwik:
Add enable service and plugin into local.conf for LBaaS v2

Fixes: #49


kiwik:
Small fix, let's merge this patch to check result.

Issue #76, opened 2018-03-12
Use neutron instead of neutron-legacy for ccm


mrhillsman:
@dims reported issue with q-* services being enabled vs neutron-*

http://logs.openlabtesting.org/logs/85/85/bb2d7e7e17eb1c2255f6da0ffade71959c641ea7/check/openstack-cloud-controller-manager-unittest/3a21176/job-output.txt.gz

reviewing the log we see q-agt,q-dhcp,q-l3,q-meta,q-metering,q-svc and these need to be changed to neutron-agent,neutron-dhcp,neutron-l3,neutron-metadata-agent,neutron-metering,neutron-api

PR #77, opened 2018-03-12
Fixes issue #76

mrhillsman:
Change q- enabled services to neutron-


hogepodge:
What about LBaaSv2 or Octavia? These are critical for cloud controller manager testing.

dims:
Approved

mrhillsman:
@hogepodge we have another for octavia that i know of but believe that is waiting for review #72 this one is for something @dims found checking job output

if there is no movement on #72 i can see what is going on with it?

dims:
@hogepodge we are trying to get the basic unit tests running. failing `TestRoutes` unit test in (occm), so trying to see if we can switch away from neutron-legacy to see if helps

PR #79, opened 2018-03-12
Fixes issue #76

mrhillsman:
Add stanza to create-devstack-local-conf role to support
neutron vs default neutron-legacy


dims:
Approved

PR #81, opened 2018-03-13
Disable all q- services

mrhillsman:
q- services are all still being run even though neutron-
services are being enabled to replace them


mrhillsman:
Still working to resolve issue #76

PR #119, opened 2018-04-05
Update to latest k8s-keystone-auth implementation

dims:
- authorization mode should fall back to RBAC as last resort. Node should be the first.
- webhook/policy.json has a hard coded project id. patch it up with our project id
- create an k8s-admin role and grant this role to "admin" user.


mrhillsman:
Approved

PR #120, opened 2018-04-05
re-use an existing role to avoid creating a new one

dims:
looks like in vexxhost we don't have privileges to create a role, plus the hassle of cleaning up, so just reuse an existing role


mrhillsman:
Approved

PR #124, opened 2018-04-06
Update host name and ip address for better e2e testing

dims:
We need to set the hostname and ip to what's known to the openstack environment


mrhillsman:
Approved

PR #127, opened 2018-04-06
protect regex brackets for specifying just conformance tests

dims:


dims:
Approved

(self-approve for tiny change)

PR #130, opened 2018-04-07
[e2e] set the node ip address since we are using external cloud provider

dims:


dims:
Approved

(self approval, trying to get e2e tests working in the weekend)

PR #131, opened 2018-04-08
[e2e] external cloud controller manager should remove node taint

dims:
There's a bunch of things that need to line up for the taint to be
automatically removed by the openstack external cloud controller. We
should have it all lined up now. So let's remove the explicit removal
here. Note that by removing the node taint by hand, we are short
circuiting some of the code in node_controller that is now run inside
the external cloud controller binary. So give that a chance to run:
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/cloud/node_controller.go#L291


dims:
Approved

PR #132, opened 2018-04-08
[e2e] oops add back hack for rbac rules

dims:
at one point we removed these (while testing e2e against old openstack
provider). Now that we are using the new external cloud provider, add
these back.


dims:
Approved

PR #133, opened 2018-04-08
Some refactor for cloud-provider-openstack-acceptance-test-e2e-conformance

kiwik:
- Use export-vexxhost-openrc role to get vexxhost credential
- Remove post.yaml
- Enable k8s cluster DNS

Partial-Bug: theopenlab/openlab#45

PR #134, opened 2018-04-08
[e2e] enable cluster dns for tests to work

dims:
last 2 failures in e2e tests seem to indicate we need the kube-dns installed and running


dims:
oops. looks like already dine in #133

PR #137, opened 2018-04-10
Use public external network for external LB scenario

dims:
"openstack network list" returns 2 networks in vexxhost, use the one with name `public`

### DIFF
--- a/playbooks/cloud-provider-openstack-acceptance-test-lb-octavia/run.yaml
+++ b/playbooks/cloud-provider-openstack-acceptance-test-lb-octavia/run.yaml
@@ -25,7 +25,7 @@
           username = $OS_USERNAME
           region = $OS_REGION_NAME
           [LoadBalancer]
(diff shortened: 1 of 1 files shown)


mrhillsman:
Approved

kiwik:
I think it is a bug of backend public cloud, see my test http://paste.openstack.org/show/718890/ , in one minute some LF-* networks appear and disappear, these LF-* networks looks like some internal network, should not be shown to end-user. It is easy to reproduce, please run "network list" in your console, repeat several times.

Let us to wait until they fix these, then close the PR.

mnaser:
This has been resolved, there is no need to do anything, only the public will be listed from now on.

mrhillsman:
@kiwik please confirm

kiwik:
thank you @mnaser and @mrhillsman, I have double-check it, looks good now, do not change anything in OpenLab side, let's close the PR and issue.

PR #142, opened 2018-04-12
Support enabling Zun on devstack

hongbin:
Gophercloud will add support for OpenStack Zun and it is using
OpenLab for testing. Therefore, we need to enable Zun in
devstack jobs. The original request comes from this PR:

  https://github.com/gophercloud/gophercloud/pull/858

### DIFF
--- a/roles/create-devstack-local-conf/tasks/main.yml
+++ b/roles/create-devstack-local-conf/tasks/main.yml
@@ -130,3 +130,22 @@
   environment: '{{ zuul | zuul_legacy_vars }}'
   when:
     - '"fwaas-v2" in enable_services'
+
+- name: create devstack local conf with zun enabled
+  shell:
+    cmd: |
+      set -e
+      set -x
+      cat << EOF >> /tmp/dg-local.conf
+      enable_service zun-api,zun-compute,zun-wsproxy
+      enable_plugin devstack-plugin-container https://git.openstack.org/openstack/devstack-plugin-container
+      KURYR_CAPABILITY_SCOPE=global
+      KURYR_ETCD_PORT=2379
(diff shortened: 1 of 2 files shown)


jtopjian:
@hongbin I realize this is out of scope of this PR, but Gophercloud has also enabled support for Zaqar, Senlin, and Octavia recently. If possible, we'd love to see testing enabled for those services, too. :)

hongbin:
@jtopjian I am not familiar with those services, but I can ping their contributors to see if they interest in working on it.

mrhillsman:
@jtopjian can you create an issue in this repo so we can address that and track it

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -130,3 +130,23 @@
...
+    cmd: |
+      set -e
+      set -x
+      cat << EOF >> /tmp/dg-local.conf
+      enable_service zun-api,zun-compute,zun-wsproxy
+      enable_plugin devstack-plugin-container https://git.openstack.org/openstack/devstack-plugin-container
+      KURYR_CAPABILITY_SCOPE=global
+      KURYR_ETCD_PORT=2379
+      enable_plugin kuryr-libnetwork https://git.openstack.org/openstack/kuryr-libnetwork
+      LIBS_FROM_GIT="python-zunclient"
```
dims: want to append if it is already set?

hongbin: Actually, I would remove it since python-zunclient is irrelevant

mrhillsman:
@hongbin @dims @liu-sheng @animationzl @kiwik merging but will revert should an error occur due to merge

mrhillsman:
Approved

PR #145, opened 2018-04-12
Support enabling Zun on devstack

hongbin:
Gophercloud will add support for OpenStack Zun and it is using
OpenLab for testing. Therefore, we need to enable Zun in
devstack jobs. The original request comes from this PR:

  https://github.com/gophercloud/gophercloud/pull/858


mrhillsman:
Approved

PR #161, opened 2018-04-23
Do refactor for Ansible jobs to keep struct consistent

kiwik:
- Add specific name attribue for all of task
- Replace OS_TENANT_* to OS_PROJECT_*
- Use consist way to get API_HOST_IP

Closes theopenlab/openlab#52

### DIFF
--- a/playbooks/cloud-provider-openstack-unittest-nested/run.yaml
+++ b/playbooks/cloud-provider-openstack-unittest-nested/run.yaml
@@ -7,7 +7,8 @@
     - install-devstack
     - create-virtual-machine
   tasks:
-    - shell:
+    - name: Run unit tests with cloud-provider-openstack (nested)
+      shell:
         cmd: |
           set -e
           set -x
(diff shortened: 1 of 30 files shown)


playbooks/cloud-provider-openstack-unittest-nested/run.yaml
```diff
@@ -1,5 +1,6 @@
 # Deprecated (Kubernetes nested scenario)
-- hosts: all
+- name: Run unit tests with cloud-provider-openstack (nested)
+   hosts: all
```
animationzl: A extra space ahead 'hosts' key

liu-sheng:
yes, the check job has found that error. also not sure if need to name all the tasks, since the ansible-lint also saide "ANSIBLE0011 All tasks should be named", which I remember in openstack official repo, there also has refactor change to add names for all the tasks.

playbooks/cloud-provider-openstack-unittest-nested/run.yaml
kiwik: Good catch, thank you.

kiwik:
@liu-sheng thank you to provider these background info, I plan to add task name in next patch.

kiwik:
recheck

Issue #162, opened 2018-04-24
loadbalancer issues with k8s testing using lbaasv2


mrhillsman:
Using octavia appears to be efficient and setting use-octavia=True to default is non-destructive as lbaasv2 endpoint is the fallback

PR #167, opened 2018-04-24
Set use-octavia to true

mrhillsman:
Signed-off-by: Melvin Hillsman <codebauss@gmail.com>

### DIFF
--- a/playbooks/cloud-provider-openstack-acceptance-test-lb-octavia/run.yaml
+++ b/playbooks/cloud-provider-openstack-acceptance-test-lb-octavia/run.yaml
@@ -28,6 +28,7 @@
           username = $OS_USERNAME
           region = $OS_REGION_NAME
           [LoadBalancer]
+          use-octavia = true
           floating-network-id = $(openstack network list --external -f value -c ID | head -n 1)
           subnet-id = $(openstack network list --internal -f value -c Subnets | head -n 1)
           [BlockStorage]


mrhillsman:
Approved

mrhillsman:
Addresses issue #162 

We should be using Octavia by default if possible due to issues with LBaaSv2 inconsistencies.

mrhillsman:
@liu-sheng @animationzl @kiwik

dims:
@liu-sheng the builds are already very flaky. how bad can it get with this change :) merging

liu-sheng:
the "use-octavia=True" option make the lbaas resource request passed to the octavia service endpoint, also need to change the resource cleanup(https://github.com/theopenlab/openlab-zuul-jobs/blob/master/playbooks/cloud-provider-openstack-acceptance-test-lb-octavia/post.yaml) to use OSC commands than neutron CLI commands. will make a corresponding PR to change that.

In my own testing, I have also add the "use-octavia=True" option and changed to use OSC commands in cleanup. but problems remains. please see: https://github.com/liu-sheng/cloud-provider-openstack/pull/12  The lbaas job will success  sometimes. but sometimes the loadbalancer created by the job may become in `ERROR` state and even cannot be deleted. seems there are some problems with Vexxhost cloud, it's hard to find the detailed reason since we cannot get the related log of vexxhost services.

Hi @mrhillsman， Ruichen has told me that you will ask the vexxhost operators about the services stability of vexxhost services ? any feadback ?

Thanks!

kiwik:
Related-Bug: https://github.com/theopenlab/openlab/issues/50

PR #169, opened 2018-04-25
Change the lbaas resource cleanup to use OSC command

liu-sheng:
This change also enlarge the waitng time of lbaas service deployment.


kiwik:
Partial-Fix: https://github.com/theopenlab/openlab-zuul-jobs/pull/167

PR #175, opened 2018-04-26
Change cinder standalone jobs to use general ubuntu-xenial node

liu-sheng:
This change can save the ubuntu-xenial-vexxhost node consumption.


liu-sheng:
@animationzl flexvolume job already use ubuntu-xenial node :)

kiwik:
@liu-sheng Ah, my fault, LTGM.

PR #176, opened 2018-04-26
Do refactor for install-golang role

kiwik:
- user can only specify version to install golang in task role
- print installed golang version into log
- remove usless role 'install-k8s-jobs-dependences'

Closes theopenlab/openlab#56


mrhillsman:
Approved

PR #200, opened 2018-05-03
add test cloud-provider-openstack-format

dixudx:
Add a new test to
* run gofmt checking
* run golint
* checking normal typos

/cc @dims


zuul.d/jobs.yaml
```diff
@@ -256,6 +256,13 @@
     secrets:
       - vexxhost_credentials
 
+- job:
+    name: cloud-provider-openstack-format
+    parent: cloud-provider-openstack-test
```
liu-sheng: The parent cloud-provider-openstack-test job use Vexxhost nodes to run, we'd better to change use general `ubuntu-xenial` nodeset for this jobs, because the amount of Vexxhost resources is limited. and we have many others K8S+OpenStack testing cases need Vexxhost nodes.

animationzl:
recheck

dims:
Approved

PR #201, opened 2018-05-03
Move gophercloud jobs to gophercloud repo

yongfengdu:
This is to remove related jobs from theopenlab repo

Fixes theopenlab/openlab#17

Related-bug: gophercloud/gophercloud#974


yongfengdu:
recheck

PR #213, opened 2018-05-09
Using cascading delete

mnaser:
The OpenStack CLI supports doing cascading deletes directly so avoid a bunch of loops and delete directly.


liu-sheng:
@mnaser thanks for this change

PR #216, opened 2018-05-15
install golang first for cloud-provider-openstack-format

dixudx:
Golang has to be installed first to make sure `gofmt` existed.

This is the root cause that causing kubernetes/cloud-provider-openstack#154 failing.

/cc @dims


liu-sheng:
don't need to install golang again, it just because the `environment: '{{ golang_env }}'` was missing in the job definition,  I have changed this job to get rid of some redundant steps, see: https://github.com/theopenlab/openlab-zuul-jobs/pull/217

dixudx:
@liu-sheng Thanks for your help.

Issue #143, opened 2018-04-12
Add Support for New Gophercloud Services


jtopjian:
Gophercloud has recently added support for:

* Zaqar / Messaging
* Senlin / Clustering
* Octavia / LBaaS

We'd like to have OpenLab be able to run the acceptance tests for these services.

mrhillsman:
@kiwik @animationzl @liu-sheng

mrhillsman:
related theopenlab/openlab/issues/17

If gophercloud does not require any secrets we need to move to gophercloud repo to empower gophercloud community to manage their jobs

jtopjian:
That would be cool :)

I've been working on cleaning up the Gophercloud acceptance tests in order to enable more tests to run: https://github.com/gophercloud/gophercloud/issues/818

Enabling/disabling tests ourselves would certainly help us as well as relieve some responsibility on your end :)

mrhillsman:
@jtopjian will work to ensure this happens sooner than later ;)

mrhillsman:
@yihleong can we get someone looking into assisting here

hongbin:
I take the Octavia part. @tengqm anyone from your team volunteers for the Senlin part. @openstacker anyone from your team volunteers for the Zaqar part?

hongbin:
@tengqm the original request is from gophercloud maintainers who wanted to run acceptance tests in their CI against Senlin devstack deployment. This repo contains the configuration of gophercloud jobs. The requirement is to enable Senlin in gophercloud jobs. See here as an example: https://github.com/theopenlab/openlab-zuul-jobs/pull/147

tengqm:
@hongbin Thanks for the hints. I'm still confused by the work items. Refer to this one: https://github.com/gophercloud/gophercloud/issues/823 There are not enough concepts merged to do any useful functional/integration tests for Senlin. 

Do you (or others) mean that we can enable Senlin on TheOpenLib in a separate thread? I'm more than happy to see this happen. However, I do not want to introduce changes that may break the existing test jobs.

hongbin:
@tengqm my understanding is:

* This repo contains the configuration of the gophercloud that basically stand up an openstack deployment using devstack
* The actual test cases are living at somewhere else (possibly in gophercloud repo).

If you or someone else wants to enable Senlin in the CI, the first step is to submit a PR to this repo to configure devstack to install Senlin. Then, you are ready to write test cases for Senlin.

jtopjian:
[...]

PR #227, opened 2018-06-02
Add Octavia devstack configuration

hongbin:
There is a lbaas devstack configuration that can enable Octavia
through neutron-lbaas. However, the neutron-lbaas is deprecated
so we need a new task for enabling Octavia as a standalone service.

https://github.com/theopenlab/openlab-zuul-jobs/issues/143


roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -100,6 +100,22 @@
   when:
     - '"lbaas" in enable_services'
 
+- name: create devstack local conf with octavia enabled
```
liu-sheng: We have added the octavia in above section(L86~L101), which add both neutron-lbaas and octavia,  do you think we need a separated `octavia` support ?

kiwik: I'm ok to add new 'octavia' service and keep 'lbaas' service still be in here, because OpenLab is a platform, should support different scenairo to different SDK/Tool, even if OpenStack community have deprecated neturon-lbaas in latest release, but still have some SDK/Tools use old release and build logic on neutron-lbaas API, like: cloud-provider-openstack. https://github.com/kubernetes/cloud-provider-openstack/blob/2bbd88dd806e63b2311898dc01b2cafe279efc2e/docs/cloud-providers.md#load-balancer

roles/install-devstack/tasks/main.yml
```diff
@@ -45,6 +45,9 @@
       export PROJECTS="openstack/devstack-plugin-container $PROJECTS"
       export PROJECTS="openstack/kuryr-libnetwork $PROJECTS"
       export PROJECTS="openstack/zun openstack/zun-tempest-plugin $PROJECTS"
+      export PROJECTS="openstack/diskimage-builder $PROJECTS"
+      export PROJECTS="openstack/tripleo-image-elements $PROJECTS"
+      export PROJECTS="openstack/octavia openstack/python-octaviaclient $PROJECTS"
```
kiwik: tripleo-image-element and octavia have been in PROJECTS, we only need add diskimage-builder and python-octaviaclient

https://github.com/openstack-infra/devstack-gate/blob/master/devstack-vm-gate-wrap.sh#L105-L119

roles/create-devstack-local-conf/tasks/main.yml
hongbin: Yes, Chen Rui made the point.

kiwik:
Hi @hongbin , we are doing an important refactor for zuul jobs, I prefer to hold this until refactor merged, then you have to rebase the master, I think refactor will be landed in tomorrow.

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -114,6 +114,21 @@
...
+      set -e
+      set -x
+      cat << EOF >> /tmp/dg-local.conf
+      enable_service octavia,o-cw,o-hm,o-hk,o-api
+      enable_plugin octavia https://git.openstack.org/openstack/octavia
+      enable_plugin barbican https://git.openstack.org/openstack/barbican
+      EOF
+    executable: /bin/bash
+    chdir: '{{ ansible_user_dir }}/workspace'
+  environment: '{{ zuul | zuul_legacy_vars }}'
```
kiwik: OK, @hongbin , you can rebase master now, maybe exist some merge conflict should be addressed, and 128 line and 129 line can be removed now.

PR #232, opened 2018-06-07
Upload multiple junit files to avoid test timeout

kiwik:
- use "--junit=/path-to/junit*.xml" to upload junit files
- change conformance test timeout from 120m to 150m
- avoid to echo gcp account info

Related-Issue: theopenlab/openlab#64


mrhillsman:
Approved

PR #235, opened 2018-06-08
Fix devstack installation in N, O, P due to Zun project

liu-sheng:
- Zun do not have release before Pike
- Need to add python-zunclient in PROJECTS environment vars for devstack


kiwik:
recheck

PR #239, opened 2018-06-15
Set datastore version for terraform trove job by querying from devstack

liu-sheng:
For # theopenlab/openlab/issues/68


liu-sheng:
The datastore version issue seems fixed, but the job still have problem, looks like issues with test case itself.
http://80.158.23.49/logs/1/1/453d6f37c00688799840b7a76a031f602eda6c37/recheck-trove/terraform-provider-openstack-acceptance-test-trove/6f99360/job-output.txt.gz

PR #242, opened 2018-06-19
Add install-k8s role

mrhillsman:
These two jobs are missing install-k8s role and are failing


liu-sheng:
sorry, seems my refactor patch cause this problem

mrhillsman:
@liu-sheng no problem, yes, i believe that is the culprit

PR #243, opened 2018-06-19
Add missing golang configuration for format job

liu-sheng:


mrhillsman:
Approved

mrhillsman:
@animationzl if we are using this role ( move-k8s-repo-to-k8s-specific-dir ) we need to also have install-k8s role above it unless I am missing something

mrhillsman:
@liu-sheng please see comment

liu-sheng:
@mrhillsman yes, sure, the move-* role is for change the path from `github/kubernetes/kubernetes` to `k8s.io/kubernetes`, the install-k8s role is for building k8s and starting up.

liu-sheng:
@mrhillsman is this format job need k8s binaries  ? now I only found it missing the `gofmt`

liu-sheng:
@mrhillsman yes I remember there is an issue occurred about the `k8s.io/kubernetes` path when we missed the move-* role, so it is not unnecessary

PR #244, opened 2018-06-19
Use q-* to enable neutron when FWaaS enabled

kiwik:
FWaaS do not switch to neutron-* devstack options totally,
we should be use q-* to enable neutron services for it in order
to avoid deploying error. Only add those enable logic into FWaaS v1
and v2 dg-local.conf to avoid impact main config path.

Close theopenlab/openlab#70

### DIFF
--- a/roles/create-devstack-local-conf/tasks/main.yml
+++ b/roles/create-devstack-local-conf/tasks/main.yml
@@ -122,7 +122,8 @@
       set -e
       set -x
       cat << EOF >> /tmp/dg-local.conf
-      enable_service neutron-fwaas-v1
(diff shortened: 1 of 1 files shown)


roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -122,7 +122,7 @@
       set -e
       set -x
       cat << EOF >> /tmp/dg-local.conf
-      enable_service neutron-fwaas-v1
+      enable_service q-agt q-svc q-dhcp q-l3 q-meta q-metering q-fwaas-v1
```
liu-sheng: is there conflict with the task `enable neutron services by new services names` in L14~L25 ?

mrhillsman:
Approved

kiwik:
@liu-sheng seems your link wrong.

PR #245, opened 2018-06-20
Add ansible functional testing jobs against shade or openstacksdk

liu-sheng:
This change add 3 jobs for running functional tests in shade or openstacksdk with different versions of ansible:

- openstacksdk-ansible-devel-functional-devstack
- openstacksdk-ansible-stable-2.6-functional-devstack
- shade-ansible-stable-2.5-functional-devstack


emonty:
Awesome!

Is it possible to just configure the openlab zuul to load jobs from the openstacksdk repo? I see that these have ```parent: init-test``` - which is different from the base job in the openstacksdk repo. Mainly it would just be neat if we can use this as an opportunity to figure out sharing job definitions for a repo across 2 zuuls with different base jobs ...

If not - or if it's too hard - let's not block progress on it - we can always circle back around and tackle the shared job thing later.

playbooks/openstacksdk-ansible-functional-devstack/run.yaml
```diff
@@ -0,0 +1,19 @@
+- hosts: all
+  become: yes
+  tasks:
+    - name: Clone opnstacksdk
+      git:
+        repo: http://git.openstack.org/openstack/openstacksdk
```
emonty: We never want to clone the repo directly from git.o.o - zuul should be preparing the git repo state and then it shoud be pushed to the node as part of the base job.

This will result in the change in question not being the thing that is tested.

emonty: To ensure that this happens - we need to add openstack/openstacksdk and openstack-infra/shade to the openlab tenant, and then add each of them to the required-projects list of the appropriate jobs.

playbooks/shade-ansible-functional-devstack/run.yaml
```diff
@@ -0,0 +1,19 @@
+- hosts: all
+  become: yes
+  tasks:
+    - name: Clone shade
+      git:
+        repo: http://git.openstack.org/openstack-infra/shade
```
emonty: Same here.

zuul.d/jobs.yaml
```diff
@@ -338,3 +338,36 @@
...
+- job:
+    name: openstacksdk-ansible-devel-functional-devstack
+    parent: init-test
+    description: |
+      Run openstacksdk ansible functional tests against a master devstack
+      using git devel branch version of ansible.
+    run: playbooks/openstacksdk-ansible-functional-devstack/run.yaml
+    required-projects:
+      - name: ansible/ansible
+        override-checkout: devel
```
emonty: Here we need to add openstack/openstacksdk

zuul.d/jobs.yaml
```diff
@@ -338,3 +338,36 @@
...
+- job:
+    name: openstacksdk-ansible-stable-2.6-functional-devstack
+    parent: init-test
+    description: |
+      Run openstacksdk ansible functional tests against a master devstack
+      using git stable-2.6 branch version of ansible.
+    run: playbooks/openstacksdk-ansible-functional-devstack/run.yaml
+    required-projects:
+      - name: ansible/ansible
+        override-checkout: stable-2.6
```
emonty: and here

zuul.d/jobs.yaml
```diff
@@ -338,3 +338,36 @@
...
+- job:
+    name: shade-ansible-stable-2.5-functional-devstack
+    parent: init-test
+    description: |
+      Run shade ansible functional tests against a master devstack
+      using git stable-2.5 branch version of ansible.
+    run: playbooks/shade-ansible-functional-devstack/run.yaml
+    required-projects:
+      - name: ansible/ansible
+        override-checkout: stable-2.5
```
emonty: and then here add openstack-infra/shade

playbooks/openstacksdk-ansible-functional-devstack/run.yaml
liu-sheng: it just also because the `openstacksdk` and `shade` have themselves job definitions in `.zuul.yaml` which have a dependencies different with openlab's job structure. if we add them in required-projects, zuul will also try to load the jobs definitions from `openstacksdk` and `shade` and will cause issue. that will cause errors. if we want to support this, we need to import the full dependencies chain of openstacksdk and shade jobs into openlab, that need to do many refactors.

playbooks/shade-ansible-functional-devstack/run.yaml
liu-sheng: ditto

zuul.d/jobs.yaml
liu-sheng: ditto

zuul.d/jobs.yaml
liu-sheng: ditto

liu-sheng:
@emonty Hi, thanks for review this ! yes I had thought about to reuse the jobs in openstacksdk and shade, but the jobs have a dependencies chain which is different with current openlab jobs, if we want to do many refactors to introduce all the dependencies projects into opanlab, and may also need to refactor existing jobs of openlab.

playbooks/openstacksdk-ansible-functional-devstack/run.yaml
emonty: Ah - so, that's ok. There is a feature you can use in the zuul main.yaml to tell it to not load any zuul config objects from the repositories:

https://github.com/openstack-infra/project-config/blob/master/zuul/main.yaml#L1428-L1431

Is an example of us using it for ansible/ansible itself (which you might also want to do for the ansible/ansible role - because in the future we will be adding a .zuul.yaml to ansible/ansible itself. ```include: []``` tells zuul "don't load any config objects from this repo" - and allows safely adding shade and openstacksdk to required-projects without the job dependency problem you mention.

I do think it would be cool for us to rework some things to allow loading the jobs too - but I think that's much work - and as you say could impact existing jobs, so let's talk about it later.

liu-sheng: That is cool, thanks for teaching me 😄 , Ihave tested, it works, will update this PR.

liu-sheng:
Hi @emonty ,

Is this looks OK now, if it looks ok, cloud you please help to install the openlab-ci github app for ansible/ansible repo ?
https://github.com/apps/theopenlab-ci

Thanks.

zuul.d/pipelines.yaml
```diff
@@ -689,3 +689,41 @@
...
+    name: third-party-check
+    description: |
+      Newly uploaded patchsets to projects that are external to OpenStack
+      enter this pipeline to receive an initial +/-1 Verified vote.
+    success-message: Build succeeded (third-party-check pipeline).
+    # TODO(mordred) We should write a document for non-OpenStack developers
+    failure-message: |
+      Build failed (third-party-check pipeline) integration testing with
+      OpenStack. For information on how to proceed, see
+      http://docs.openstack.org/infra/manual/developers.html#automated-testing
```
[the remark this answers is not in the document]
liu-sheng: done, maybe just remove this, there isn't a proper openlab link for here

zuul.d/pipelines.yaml
```diff
@@ -689,3 +689,41 @@
...
+      github:
+        status: 'success'
+      mysql:
+    failure:
+      github:
+        status: 'failure'
+      mysql:
+    # Don't report merge-failures to github
+    merge-failure:
+      mysql:
```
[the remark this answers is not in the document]
liu-sheng: done

liu-sheng:
please just ignore the job check failure, it is because the github app didn't be installed

zuul.d/pipelines.yaml
```diff
@@ -689,3 +689,34 @@
       github:
         status: 'failure'
       mysql:
+
+- pipeline:
+    name: third-party-check
```
kiwik: I think we can re-use `check` pipeline at here, this config is same with `check` pipeline.

zuul.d/projects.yaml
```diff
@@ -0,0 +1,33 @@
+- project:
+    check:
+      jobs:
+        - openlab-zuul-jobs-check
+
+- project:
+    name: ansible/ansible
+    third-party-check:
```
kiwik: We can re-use `check`pipeline at here, and if we want to support more OpenStack versions, `recheck-*` pipelines are ready to be used, then we can keep consistent config for ansible, gophercloud and terraform.

liu-sheng: OK, done

liu-sheng:
sorry, I have  made a mess of this changes :(, will open a new PR for this

PR #249, opened 2018-06-20
Update dockerhub secret data

mrhillsman:
Pushing images to dockerhub failing due to incorrect
username/password combination. Could possibly be bad
format in secrets file so this updates format and data.


dims:
Approved

PR #260, opened 2018-07-06
Switch to use new OpenLab CI: update log server key

liu-sheng:


liu-sheng:
Please don't merge this until we have done switch

PR #272, opened 2018-07-26
Add support for running packer functionality test against devstack

liu-sheng:


zuul.d/projects.yaml
```diff
@@ -46,3 +46,7 @@
         - packer-1.2.5-functional-orange
         - packer-1.2.5-functional-telefonica
         - packer-1.2.5-functional-huaweicloud
+    check:
+      jobs:
+        - packer-master-functional-devstack
+        - packer-1.2.5-functional-devstack
```
kiwik: We can put `packer-1.2.5-functional-devstack` in periodic pipeline, and use `packer-master-functional-devstack` to check PR changes.

PR #280, opened 2018-08-08
Unset AZ env and correct the orange auth_url to fix docker machine jobs

liu-sheng:
- The `OS_AVAILABILITY_ZONE` env will be set in `export-cloud-openrc` in opentelekomcloud, which will cause error in docker machine jobs, this PR unset the variable in docker machine job
- The `auth_url`  of orange cloud in secrets don't include a "/v3" postfix, which will cause error in docker machine testing cases. This PR correct it.


liu-sheng:
recheck

PR #281, opened 2018-08-08
Add support for running docker machine tests against devstack

liu-sheng:


zuul.d/jobs.yaml
```diff
@@ -665,3 +665,15 @@
     override-checkout: v0.15.0
     secrets:
       - huaweicloud_credentials
+
+- job:
+    name: docker-machine-0.15.0-functional-devstack-mitaka
+    parent: init-test
+    description: |
+      Test image building functionality of Docker machine of 0.15.0 version against devstack of Mitaka.
+    run: playbooks/docker-machine-functional-devstack-mitaka/run.yaml
```
kiwik: We can reuse the job for all of OpenStack releases, so please do not append `mitaka` at job name, like packer job.

playbooks/docker-machine-functional-devstack-mitaka/run.yaml
```diff
@@ -0,0 +1,59 @@
...
+          cp ./bin/docker-machine /usr/local/bin/
+          docker-machine version
+
+          export OS_FLAVOR_ID=11
+          export OS_IMAGE_NAME="Ubuntu 16.04 Server 64bit"
+          export OS_DOMAIN_ID=default
+          export OS_SSH_USER=ubuntu
+          export OS_NETWORK_NAME=private
+          export OS_FLOATINGIP_POOL=public
+          openstack security group rule create --proto tcp --dst-port 22 default
```
kiwik: Need to do some command options compatible for mitaka and other releases so that the job can be used for all OpenStack releases.

playbooks/docker-machine-functional-devstack-mitaka/run.yaml
```diff
@@ -0,0 +1,59 @@
...
+          apt-key adv --keyserver hkp://pgp.mit.edu:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D || true
+          apt-add-repository 'deb http://apt.dockerproject.org/repo ubuntu-trusty main'
+          apt-get update
+          apt-cache policy docker-engine
+          apt-get install -y docker-engine=1.12.6-0~ubuntu-trusty
+          cat /lib/systemd/system/docker.service
+          sed -r -i "s|(ExecStart)=(.+)|\1=\2 --iptables=false|" /lib/systemd/system/docker.service
+          cat /lib/systemd/system/docker.service
+          service docker restart
+          ifconfig -a
```
kiwik: Whether we can do some refactor in order to put install-docker into a role?

zuul.d/jobs.yaml
liu-sheng: because of the auth url problem this job cannot run with other releases of OpenStack, and also  the docker installation packages are specific for ubuntu trusty.

playbooks/docker-machine-functional-devstack-mitaka/run.yaml
liu-sheng: as this job can  only run with Mitaka, because of the auth url issue. maybe this is not a generic job definition for other OpenStack releases.  Also, the docker packages are specific for `ubuntu-trusty`

playbooks/docker-machine-functional-devstack-mitaka/run.yaml
liu-sheng: not very sure, but here it is specific for ubuntu trusty. and also in ubuntu 14.04. we can only use `service xxx restart`, the `systemctl` command not supportted.

zuul.d/jobs.yaml
kiwik: We should show the result even if docker-machine can not work with these OpenStack release.

playbooks/docker-machine-functional-devstack-mitaka/run.yaml
kiwik: We should not need to install specific docker version, let package manage tools to handle these, just install like
```
apt-get install -y docker-engine
```

zuul.d/jobs.yaml
```diff
@@ -707,3 +707,64 @@
...
+        OS_BRANCH: stable/mitaka
+    override-checkout: v0.15.0
+    nodeset: ubuntu-trusty
+
+- job:
+    name: docker-machine-0.15.0-functional-devstack-newton
+    parent: init-test
+    description: |
+      Test image building functionality of Docker machine of 0.15.0 version against
+      devstack of Mitaka.
```
kiwik: s/Mitaka/Newton/

zuul.d/jobs.yaml
```diff
@@ -707,3 +707,64 @@
...
+      global_env:
+        OS_BRANCH: stable/newton
+    override-checkout: v0.15.0
+
+- job:
+    name: docker-machine-0.15.0-functional-devstack-ocata
+    parent: init-test
+    description: |
+      Test image building functionality of Docker machine of 0.15.0 version against
+      devstack of Mitaka.
```
kiwik: s/Mitaka/Ocata/

zuul.d/jobs.yaml
```diff
@@ -707,3 +707,64 @@
...
+      global_env:
+        OS_BRANCH: stable/ocata
+    override-checkout: v0.15.0
+
+- job:
+    name: docker-machine-0.15.0-functional-devstack-pike
+    parent: init-test
+    description: |
+      Test image building functionality of Docker machine of 0.15.0 version against
+      devstack of Mitaka.
```
kiwik: s/Mitaka/Pike/

zuul.d/jobs.yaml
```diff
@@ -707,3 +707,64 @@
...
+      global_env:
+        OS_BRANCH: stable/pike
+    override-checkout: v0.15.0
+
+- job:
+    name: docker-machine-0.15.0-functional-devstack-queens
+    parent: init-test
+    description: |
+      Test image building functionality of Docker machine of 0.15.0 version against
+      devstack of Mitaka.
```
kiwik: s/Mitaka/Queens/

PR #288, opened 2018-08-15
Fix multiple default security group conflicts issue

liu-sheng:


kiwik:
Close theopenlab/openlab#74

PR #300, opened 2018-08-31
Use theopenlab/ansible instead of ansible/ansible

kiwik:
Github APP is not installed into offical ansible project yet,
avoid to hit Github API rate limits, config openlab forked ansible.


kiwik:
recheck

PR #305, opened 2018-08-31
Update run.yaml

zhongjun2:


liu-sheng:
switch to use allinone openlab CI, close this PR

PR #308, opened 2018-09-04
osb checker use huaweicloud nodeset

kiwik:
Related-Bug: theopenlab/openlab#75


kiwik:
should check whether osb-checker jobs must run on huaweicloud nodes.

kiwik:
@liu-sheng according to current testing status, `osb-checker` do have to run on huaweicloud node, so I remove node set setting, I just add some comments in jobs.yaml to make it readable, please help to review when you have time, thank you.

PR #318, opened 2018-09-12
Fix packer-1.2.5-functional-orange failing

kiwik:
We test packer with flavor t2.micro against Orange cloud,
vm can be launched and active, but guest OS and sshd process
launch slowly, that cause packer check ssh connection timeout.

Closes: theopenlab/openlab#80

### DIFF
--- a/playbooks/packer-functional-public-clouds/run.yaml
+++ b/playbooks/packer-functional-public-clouds/run.yaml
@@ -10,7 +10,9 @@
           set -ex
           set -o pipefail
           pip install python-openstackclient
-          flavor=$(openstack flavor list -f value -c ID -c RAM -c VCPUs --sort-column RAM --sort-column VCPUs | head -n 1 | awk '{print $1}')
+          # Using second smallest flavor, because first one might cause testing failed.
+          # See https://github.com/theopenlab/openlab/issues/80
+          flavor=$(openstack flavor list -f value -c ID -c RAM -c VCPUs --sort-column RAM --sort-column VCPUs | awk 'NR==2{print $1}')
           network=$(openstack network show openlab-jobs-net -f value -c id)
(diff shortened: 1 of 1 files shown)


playbooks/packer-functional-public-clouds/run.yaml
```diff
@@ -10,7 +10,9 @@
           set -ex
           set -o pipefail
           pip install python-openstackclient
-          flavor=$(openstack flavor list -f value -c ID -c RAM -c VCPUs --sort-column RAM --sort-column VCPUs | head -n 1 | awk '{print $1}')
+          # Using second smallest flavor, because first one might cause testing failed.
+          # See https://github.com/theopenlab/openlab/issues/80
+          flavor=$(openstack flavor list -f value -c ID -c RAM -c VCPUs --sort-column RAM --sort-column VCPUs | awk 'NR==2{print $1}')
```
liu-sheng: maybe you can try like below, it is looks a more generic way: 

```
MEDIUM_FLAVOR=$(openstack flavor list -f value -c ID -c RAM -c VCPUs --sort-column RAM --sort-column VCPUs | awk '{if($2>=2048 && $3>=2){print $1}}' | head -n 1)
```
it means the fist flavor with mem >= 2048m and vCPU >=2 :)

kiwik: I think we should't add so many `magic` code, just keep command simple, and awk $2 $3 have to match the first command order of  `-c` option, that's difficult to maintain for other developers.

liu-sheng: For different clouds, the flavors maybe different, this way select the flavor only according to mem and  vCPUs to avoid confusion. maybe just add a NOTE for this statement to explain for other developers. 
but I am also ok with your change.

mrhillsman: Why not increase the timeout?

PR #319, opened 2018-09-12
manila-provisioner job: fixed sourcing openrc params

gman0:
This PR:
* fixes sourcing of OpenRC params when creating k8s secret
* adds a sleep to wait for ServiceAccount default/default to complete


gman0:
PTAL @kiwik

PR #321, opened 2018-09-17
manila-provisioner job: set manila share type

gman0:
This PR:
* sets correct manila share type in the `StorageClass`
* adds a `sleep` in between deleting the pvc+pod and sc+secret - just in case


gman0:
PTAL @kiwik

PR #327, opened 2018-09-20
Update secrets.yaml for TLF and HuaweiCloud

kiwik:
Closes: theopenlab/openlab#85


liu-sheng:
@h00130372 it is the prefix of the AK, just to identify which AK/SK used, for convenience to read the code.

PR #337, opened 2018-10-16
Remove duplicate job define "cloud-provider-openstack-test"

kiwik:
Closes: theopenlab/openlab#88


zuul.d/jobs.yaml
```diff
@@ -238,12 +238,10 @@
 
 - job:
     name: cloud-provider-openstack-format
-    parent: golang-test
+    parent: cloud-provider-openstack-test
```
liu-sheng: Why change the parent of cloud-provider-openstack-format job to cloud-provider-openstack-test ? seems we don't need the cleanup steps in post.yaml. only need to move the k8s dir path.

liu-sheng: after discussion, let's merge this firstly, and improve the post cleanup steps in the future.

kiwik: Yes, thank you @liu-sheng

PR #343, opened 2018-10-23
Add manageiq-providers-openstack job for public clouds

liu-sheng:
For #theopenlab/openlab/issues/81

### DIFF
--- a/playbooks/manageiq-providers-openstack-test-public-clouds/run.yaml
+++ b/playbooks/manageiq-providers-openstack-test-public-clouds/run.yaml
@@ -61,13 +61,11 @@
           bin/setup
           bundle install
           cp openstack_environments.yml spec/manageiq/
-
-          pushd spec/manageiq/
-          sed -i "/admin_role =/ s|:domain_id => domain_id||"  spec/tools/environment_builders/openstack/services/identity/builder/keystone_v3.rb
-          ln -s $(realpath ../../spec/models/manageiq/providers/openstack) spec/models/manageiq/providers/openstack
-          bundle install
-          bundle exec rake db:setup
-          popd
+          set +x
+          sed -i "/setup_ems/ s|admin|$OS_USERNAME|" spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_keystone_v3_spec.rb
+          sed -i "/setup_ems/ s|5000|433|" spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_keystone_v3_spec.rb
+          sed -i "/setup_ems/ s|default|$OS_DOMAIN_ID|" spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_liberty_keystone_v3_spec.rb
+          set -x
 
           rm spec/models/manageiq/providers/openstack/cloud_manager/refresher_rhos_havana_errors_spec.rb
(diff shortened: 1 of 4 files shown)


playbooks/manageiq-providers-openstack-test-public-clouds/run.yaml
```diff
@@ -0,0 +1,77 @@
+- hosts: all
+  become: yes
+  roles:
+    - export-cloud-openrc
+  tasks:
+    - name: Run Integration tests of Docker machine against devstack
```
h00130372: please modify the task name, it's tests of manageiq not Docker
machine.

playbooks/manageiq-providers-openstack-test-public-clouds/run.yaml
```diff
@@ -0,0 +1,77 @@
...
+      shell:
```
h00130372: can we put these cmd into a role which doing manageiq developer setup?

playbooks/manageiq-providers-openstack-test-public-clouds/run.yaml
liu-sheng: done, thx

playbooks/manageiq-providers-openstack-test-public-clouds/run.yaml
```diff
@@ -0,0 +1,77 @@
...
+          - liberty_keystone_v3:
+              ip: $auth_netloc
+              password: $OS_PASSWORD
+              user: $OS_USERNAME
+          EOF
+          set -x
+
+          bin/setup
+          bundle install
+          cp openstack_environments.yml spec/manageiq/
```
kiwik:
```
          cp openstack_environments.yml spec/manageiq/

          pushd spec/manageiq/
          sed -i "/admin_role =/ s|:domain_id => domain_id||"  spec/tools/environment_builders/openstack/services/identity/builder/keystone_v3.rb
          ln -s $(realpath ../../spec/models/manageiq/providers/openstack) spec/models/manageiq/providers/openstack
          bundle install
          bundle exec rake db:setup
          popd
```
These lines build OpenStack resources in devstack, if we use public cloud, not need these.

playbooks/manageiq-providers-openstack-test-public-clouds/run.yaml
```diff
@@ -0,0 +1,77 @@
...
+          wget http://ftp.cz.debian.org/debian/pool/main/o/openssl1.0/libssl1.0.2_1.0.2o-1_amd64.deb
+          apt remove -y libssl-dev
+          dpkg -i libssl1.0-dev_1.0.2o-1_amd64.deb libssl1.0.2_1.0.2o-1_amd64.deb
+
+          apt-get install -y postgresql python-psycopg2 libpq-dev
+
+          set +x
+          # Prepare openstack_environments.yml
+          auth_netloc=$(python -c "import urlparse;print urlparse.urlparse('$OS_AUTH_URL').netloc")
+          cat << EOF >> openstack_environments.yml
```
kiwik: we should update `user`, `domain_id` and `port` in refresher_rhos_liberty_keystone_v3_spec.rb for public cloud.

playbooks/manageiq-providers-openstack-test-public-clouds/run.yaml
kiwik: +1

PR #351, opened 2018-11-01
Fix manageiq-providers-openstack-test-devstack issues

kiwik:
- openstackclient don't support --sort-column option in M, N, O, P,
  so use more simple command to handle flavor.
- sync with upstream of manageiq-providers-openstack to address
  bin/setup issue.

Closes: theopenlab/openlab#95

### DIFF
--- a/roles/clean-devstack-default-resources/tasks/main.yaml
+++ b/roles/clean-devstack-default-resources/tasks/main.yaml
@@ -17,5 +17,14 @@
       openstack subnet delete public-subnet
       openstack subnet delete private-subnet
       openstack network delete private public
-      openstack flavor delete `openstack flavor list -f value -c ID --sort-column ID | tail -n +6` || true
+      flavor_white_list="'m1.tiny' 'm1.small' 'm1.medium' 'm1.large' 'm1.xlarge'"
+      for flavor_name in `openstack flavor list -f value -c Name`
+      do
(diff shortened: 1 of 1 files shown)


roles/clean-devstack-default-resources/tasks/main.yaml
```diff
@@ -17,5 +17,14 @@
       openstack subnet delete public-subnet
       openstack subnet delete private-subnet
       openstack network delete private public
-      openstack flavor delete `openstack flavor list -f value -c ID --sort-column ID | tail -n +6` || true
+      flavor_white_list="'m1.tiny' 'm1.small' 'm1.medium' 'm1.large' 'm1.xlarge'"
+      for flavor_name in `openstack flavor list -f value -c Name`
+      do
+          if [[ "${flavor_white_list}" =~ "${flavor_name}" ]]; then
```
liu-sheng: I just tried to find a simple way:
```
flavors=`echo $(openstack flavor list -f value -c ID) | python -c "import sys; print ' '.join(list(set(sys.stdin.readline().split())-set(['1','2','3','4','5'])))"`
openstack flavor delete $flavors
```

But this is OK if you like this.

kiwik: Thank you sharing this way, but I want to keep simple and readable, I think my way is more easy to understand.

PR #354, opened 2018-11-05
Use upstream go2xunit to instead of OpenLab forked

kiwik:
- Use tebeka/go2xunit to instead of theopenlab/go2xunit
- Build file `nosetests.xslt` into jobs

Related-Bugs: theopenlab/openlab#100

### DIFF
--- a/roles/format-golang-test-results/tasks/main.yml
+++ b/roles/format-golang-test-results/tasks/main.yml
@@ -1,14 +1,24 @@
-- name: format golang test results to html
+- name: Format golang test results to XML
   shell:
     cmd: |
       set -e
       set -x
-
-      go get -u github.com/theopenlab/go2xunit
+      go get -u github.com/tebeka/go2xunit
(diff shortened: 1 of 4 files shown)


roles/format-golang-test-results/tasks/main.yml
```diff
@@ -4,11 +4,11 @@
       set -e
       set -x
 
-      go get -u github.com/theopenlab/go2xunit
+      go get -u github.com/tebeka/go2xunit
       go2xunit -input $TEST_RESULTS_TXT -output $TEST_RESULTS_XML || true
 
       apt-get install --yes xsltproc
-      wget https://raw.githubusercontent.com/theopenlab/nosetest-xunit-xslt/master/nosetests.xslt
```
liu-sheng: maybe don't need to wget ? seems there is built-in ansible vars can be used

kiwik: Thank you, will use `get_url`

PR #358, opened 2018-11-06
Fix manageiq job against Mitaka devstack

liu-sheng:
There are are some packages need to be installed for specific Ubuntu 14.04, this PR add two tasks for specific Ubuntu14.04 and Ubuntu16.04. This PR also fix the neutronclient CLI commands of master of Devstack.

Related-Bug: theopenlab/openlab#101
Related-Bug: theopenlab/openlab#81

### DIFF
--- a/playbooks/manageiq-providers-openstack-test-devstack/run.yaml
+++ b/playbooks/manageiq-providers-openstack-test-devstack/run.yaml
@@ -40,7 +40,7 @@
           # Install Ruby
           apt-get install -y build-essential zlibc zlib1g-dev ruby ruby-dev openssl libxslt-dev libxml2-dev \
               libssl-dev libreadline6 libreadline6-dev libyaml-dev libsqlite3-dev sqlite3
-          apt-get install -y postgresql python-psycopg2 libpq-dev
+          apt-get install -y postgresql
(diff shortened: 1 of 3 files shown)


playbooks/manageiq-providers-openstack-test-devstack/run.yaml
```diff
@@ -11,13 +11,50 @@
...
+        - global_env.OS_BRANCH == 'stable/mitaka'
+
+    - name: Install specifc packages and config Neutron for Ubuntu 16.04
+      shell:
+        cmd: |
+          set -ex
+          # Install Ruby
+          apt-get install -y build-essential zlibc zlib1g-dev ruby ruby-dev openssl libxslt-dev libxml2-dev \
+              libssl-dev libreadline6 libreadline6-dev libyaml-dev libsqlite3-dev sqlite3
+          apt-get install -y postgresql python-psycopg2 libpq-dev
```
kiwik: `apt-get install -y python-psycopg2 libpq-dev` command will be executed in line 83, so we should only install `postgresql` at here.

liu-sheng: thx done

PR #369, opened 2018-11-27
Clean forked project theopenlab/packer and theopenlab/machine

kiwik:
1. Use upstream projects to instead of OpenLab forked projects to
   avoid syncing with upstream manually.
2. Add UTC-0 04:00 in periodic pipeline in order to have more chance
   to trigger periodic jobs and show issue of periodic jobs

Closes: theopenlab/openlab#133

### DIFF
--- a/roles/config-packer/tasks/main.yaml
+++ b/roles/config-packer/tasks/main.yaml
@@ -7,12 +7,12 @@
     - name: install packer from source
       shell: |
         set -ex
-        mkdir -p $GOPATH/src/github.com/hashicorp/packer
-        cp -r '{{ zuul.project.src_dir }}' $GOPATH/src/github.com/hashicorp/
-        cd $GOPATH/src/github.com/hashicorp/packer
         make dev
         packer version
       environment: '{{ global_env }}'
+      args:
+        executable: /bin/bash
+        chdir: '{{ zuul.project.src_dir }}'
   when:
     - packer_version == "master"
 
--- a/zuul.d/pipelines.yaml
+++ b/zuul.d/pipelines.yaml
@@ -176,12 +176,12 @@
 
 - pipeline:
     name: periodic
-    description: Jobs in this queue are triggered on a timer. UTC-0 0:00, 8:00 and 16:00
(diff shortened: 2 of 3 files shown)


zuul.d/pipelines.yaml
```diff
@@ -176,12 +176,12 @@
 
 - pipeline:
     name: periodic
-    description: Jobs in this queue are triggered on a timer. UTC-0 0:00, 8:00 and 16:00
+    description: Jobs in this queue are triggered on a timer. UTC-0 0:00, 04:00, 08:00 and 16:00
```
liu-sheng: do we really need to run periodic jobs 4 times per day ? looks like a bit more frequency. actually, I personally think the previous 3 times per day even more frequency :)

roles/config-packer/tasks/main.yaml
```diff
@@ -7,9 +7,6 @@
     - name: install packer from source
       shell: |
         set -ex
-        mkdir -p $GOPATH/src/github.com/hashicorp/packer
-        cp -r '{{ zuul.project.src_dir }}' $GOPATH/src/github.com/hashicorp/
-        cd $GOPATH/src/github.com/hashicorp/packer
```
liu-sheng: maybe need to add a `chdir: '{{ zuul.project.src_dir }}'` if we remove the `cd` statement ?

zuul.d/pipelines.yaml
kiwik: Three times in one day 04:00, 08:00 and 16:00 , it's ok?

roles/config-packer/tasks/main.yaml
kiwik: Thank you, will update PR

zuul.d/pipelines.yaml
liu-sheng: yes, I prefer one day 04:00, 08:00 and 16:00, as it is only for us to conveniently to check the results in working time. otherwise, I think 1 time per day is enough. it is impossible the tested projects code changed many times every day.

liu-sheng:
thank you

PR #370, opened 2018-11-28
Do refactor for docker-machine jobs

kiwik:
1. Add role "config-docker-machine" to compiled and install
   docker-machine from source code.
2. Change job logic in order to putting related code together
3. Fix cp to same path issue, because we have used upstream project,
   don't need to cp forked project into official path

Related-Bug: theopenlab/openlab#133

### DIFF
--- a/roles/config-docker-machine/tasks/main.yaml
+++ b/roles/config-docker-machine/tasks/main.yaml
@@ -1,0 +1,28 @@
+---
+- name: Install and config golang
+  include_role:
+    name: config-golang
+
+- name: Compiled and install docker-machine from source
+  shell:
+    cmd: |
+      set -ex
+
+      # Install docker
+      apt-get update
+      apt-get install apt-transport-https ca-certificates
+      wget -qO- https://get.docker.com/ | sh
+
+      # NOTE: backport fix of PR 4543 and PR 4545
+      git config --global user.email 'zuul@openlab.com'
+      git config --global user.name 'openlab'
+      git cherry-pick 9ec6729
+      git cherry-pick 7909ee3
+
+      # Build docker-machine
+      make build
+      cp ./bin/docker-machine /usr/local/bin/
+      docker-machine version
+    executable: /bin/bash
+    chdir: '{{ zuul.project.src_dir }}'
+  environment: '{{ global_env }}'
--- a/playbooks/docker-machine-functional-devstack/run.yaml
+++ b/playbooks/docker-machine-functional-devstack/run.yaml
@@ -7,40 +7,16 @@
     - clone-devstack-gate-to-workspace
     - create-devstack-local-conf
     - install-devstack
-    - config-golang
+    - config-docker-machine
   tasks:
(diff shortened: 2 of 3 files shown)


roles/config-docker-machine/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+      name: config-golang
+  - name: install docker-machine from source
+    shell: |
+      set -ex
+
+      # Install bats
+      git clone https://github.com/sstephenson/bats.git '{{ ansible_user_dir }}/bats'
+      pushd '{{ ansible_user_dir }}/bats'
+      ./install.sh /usr/local
+      popd
```
liu-sheng: bats is not related with docker-machine installation, it is only for running tests.

roles/config-docker-machine/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
+---
+- name: Compiled and install docker-machine from source
+  block:
```
liu-sheng: I don't think we need `block` here

roles/config-docker-machine/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+      # Install bats
+      git clone https://github.com/sstephenson/bats.git '{{ ansible_user_dir }}/bats'
+      pushd '{{ ansible_user_dir }}/bats'
+      ./install.sh /usr/local
+      popd
+
+      # Install docker
+      apt-get update
+      apt-get install apt-transport-https ca-certificates
+      wget -qO- https://get.docker.com/ | sh
```
liu-sheng: docker installation also looks a bit unrelated with docker-machine installation,.. but OK

roles/config-docker-machine/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+      # NOTE: backport fix of PR 4543 and PR 4545
+      git config --global user.email 'zuul@openlab.com'
+      git config --global user.name 'openlab'
+      git cherry-pick 9ec6729
+      git cherry-pick 7909ee3
```
liu-sheng: maybe we can switch to use the latest release v0.16.0 and remove this workaround

kiwik: Yes, I consider that too, will do it in separated patch.

roles/config-docker-machine/tasks/main.yaml
kiwik: OK

roles/config-docker-machine/tasks/main.yaml
kiwik: Putting docker and docker-machine installation together make sense, docker-machine tests depends on it.

roles/config-docker-machine/tasks/main.yaml
kiwik: OK, will move into job logic

PR #393, opened 2018-12-19
Run npm install under osb-checker/common

liu-sheng:
As the osb-checker has been refactor, it need also to run `npm install`
under the osb-checker/common directory

Closes-Bug: #theopenlab/openlab#144


liu-sheng:
recheck

PR #398, opened 2018-12-27
Set up Devstack with manila enbaling the LVM driver

h00130372:
Modify the local.conf to set up Devstack with manila
enabling the LVM driver which operates in
driver_handles_share_services=False mode.

Related-Bug: theopenlab/openlab#142

### DIFF
--- a/roles/create-devstack-local-conf/tasks/main.yml
+++ b/roles/create-devstack-local-conf/tasks/main.yml
@@ -104,22 +104,31 @@
       set -e
       set -x
       cat << EOF >> /tmp/dg-local.conf
+      MANILA_MULTI_BACKEND=True
       MANILA_BACKEND1_CONFIG_GROUP_NAME=london
       MANILA_BACKEND2_CONFIG_GROUP_NAME=paris
       MANILA_SHARE_BACKEND1_NAME=LONDON
       MANILA_SHARE_BACKEND2_NAME=PARIS
-      MANILA_OPTGROUP_london_driver_handles_share_servers=True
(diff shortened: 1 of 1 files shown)


roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -104,22 +104,31 @@
...
-      MANILA_DEFAULT_SHARE_TYPE_EXTRA_SPECS='snapshot_support=True create_share_from_snapshot_support=True'
+
+      # This local.conf sets up Devstack with manila enabling the LVM
+      # driver which operates in driver_handles_share_services=False mode
+      MANILA_OPTGROUP_london_driver_handles_share_servers=False
+      MANILA_OPTGROUP_paris_driver_handles_share_servers=False
+      SHARE_DRIVER=manila.share.drivers.lvm.LVMShareDriver
+      MANILA_DEFAULT_SHARE_TYPE_EXTRA_SPECS='snapshot_support=True create_share_from_snapshot_support=True revert_to_snapshot_support=True mount_snapshot_support=True'
+
+      # This local.conf sets up Devstack with manila enabling the Generic
```
liu-sheng: can this commented block L120~L131 be deleted directly ?

h00130372: I prefer to leave them there, if we want to change to another mode, then we can use the options directly.

kiwik: OK, let us to run it online.

PR #401, opened 2019-01-04
Change the time of trigger of openshift + public clouds job

liu-sheng:
Close: theopenlab/openlab#149

### DIFF
--- a/zuul.d/pipelines.yaml
+++ b/zuul.d/pipelines.yaml
@@ -188,13 +188,13 @@
       mysql:
 
 - pipeline:
-    name: periodic-utc-2
-    description: Jobs in this queue are triggered on a timer. UTC-0 02:00
+    name: periodic-2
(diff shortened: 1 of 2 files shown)


zuul.d/pipelines.yaml
```diff
@@ -188,13 +188,13 @@
       mysql:
 
 - pipeline:
-    name: periodic-utc-2
-    description: Jobs in this queue are triggered on a timer. UTC-0 02:00
+    name: periodic-utc-22
```
kiwik: I think we should not associate pipeline name with trigger time, because we might add more trigger times in the future.

liu-sheng: maybe we'd better to more pipelines than one pipeline with multiple timers ? or even we add more tigger times, we can modify the pipeline name to identify more times. not sure if this is OK, but this looks more intuitive :) if don't use this name, name it "perodic-2" ?

PR #410, opened 2019-01-15
Do refactor for Ruby install

kiwik:
Build a role to install specific version of Ruby by using rbenv,
and do refactor for manageiq jobs to remove ruby installing logic
in main.yaml.

Closes: theopenlab/openlab#159


kiwik:
All-in-one env testing is ok, all-in-one job is different with regular job, that cause testing failed. This PR is ready to go.

http://80.158.7.210/logs/5/5/2168c43e9cafc6dd598eced5fe5a1cf2f21ca357/check/manageiq-providers-openstack-test-devstack/5a80008/

PR #414, opened 2019-01-21
Don't need to disable httpd_mod_wsgi for docker machine + Devstack jobs

liusheng:
After testing, we don't need to set "DISABLE_HTTPD_MOD_WSGI" to true for
docker machine + devstack jobs and only need to add the '/v3' postfix if
AUTH_URL doesn't include it. This can avoid the SSH failure problem.

Closes: theopenlab/openlab#168


liu-sheng:
please see testing results: https://github.com/liu-sheng/machine/pull/6#issuecomment-455941512

PR #420, opened 2019-01-28
Adjust csi-attacher name for new release

johscheuer:
Currently the `cloud-provider-openstack-acceptance-test-csi-cinder` job in https://github.com/kubernetes/cloud-provider-openstack fails (see: https://github.com/kubernetes/cloud-provider-openstack/pull/430 + logs: http://logs.openlabtesting.org/logs/30/430/044b09b3d0d07fcb43a678a9e30c8dfb48c68c97/cloud-provider-openstack-acceptance-test-csi-cinder/cloud-provider-openstack-acceptance-test-csi-cinder/e8e2df9/job-output.txt.gz just search for `Error response from daemon: No such image: quay.io/k8scsi/csi-attacher:canary`)

Running the `make` command like in the Zuul job creates an `csi-attacher:latest` image like defined here: https://github.com/kubernetes-csi/external-attacher/blob/master/release-tools/build.make#L68 probably it would be a good idea to pin the version of the tools (e.g. csi-attacher)?

### DIFF
--- a/playbooks/cloud-provider-openstack-acceptance-test-csi-cinder/run.yaml
+++ b/playbooks/cloud-provider-openstack-acceptance-test-csi-cinder/run.yaml
@@ -116,7 +116,7 @@
 
           git clone https://github.com/kubernetes-csi/external-attacher ${GOPATH}/src/github.com/kubernetes-csi/external-attacher
           make -C ${GOPATH}/src/github.com/kubernetes-csi/external-attacher container
-          docker tag quay.io/k8scsi/csi-attacher:canary docker.io/k8scsi/csi-attacher:latest
+          docker tag csi-attacher:latest docker.io/k8scsi/csi-attacher:latest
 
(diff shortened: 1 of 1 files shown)


dims:
@adisky @flaper87  can you please take a look?

adisky:
@johscheuer we are pinned to specific versions only 
https://github.com/theopenlab/openlab-zuul-jobs/blob/master/playbooks/cloud-provider-openstack-acceptance-test-csi-cinder/run.yaml#L132
https://github.com/kubernetes/cloud-provider-openstack/blob/master/manifests/cinder-csi-plugin/csi-attacher-cinderplugin.yaml#L33

adisky:
@johscheuer since we are pinned to 0.4.1, building and tagging here is unnecessary 
https://github.com/theopenlab/openlab-zuul-jobs/blob/master/playbooks/cloud-provider-openstack-acceptance-test-csi-cinder/run.yaml#L113-L123
removing it is a better option

johscheuer:
@adisky thanks for the fast feedback I remove the lines mentioned above

adisky:
lgtm

PR #421, opened 2019-01-29
Disable all the HuaweiCloud jobs as the account expire

liu-sheng:


liu-sheng:
abandon as we have changed the password of huaweicloud

PR #427, opened 2019-02-13
Skip ansible-lint 601 error

h00130372:
We skip ansible-lint 601 error first to make sure job
checking success.


h00130372:
fix the error instead

PR #428, opened 2019-02-13
Fix ansible-lint 601 error

h00130372:
Fix ansible-lint 601 error don't compare to literal
ture/false.


h00130372:
fix this with ansible-lint 306 in other pull request

PR #430, opened 2019-02-18
Add job for running spark funcional testing on native k8s cluster

liu-sheng:
Closes: # theopenlab/openlab#188

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -1065,3 +1065,10 @@
       inventory_name: orange
     secrets:
       - orange_credentials
+
+- job:
+    name: spark-funcional-test-k8s
+    parent: init-test
+    description: |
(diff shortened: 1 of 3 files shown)


liu-sheng:
@kiwik @h00130372 what do you think about adding a `openlab-dummy` empty project for triggering such jobs, which don't need a real project to trigger ?

zuul.d/jobs.yaml
```diff
@@ -1065,3 +1065,10 @@
       inventory_name: orange
     secrets:
       - orange_credentials
+
+- job:
+    name: spark-funcional-test-k8s
```
kiwik: The job download Java binary Spark release directly, so please use `spark-bin-*` as job name, currently all of other OpenLab jobs compile project from source code

liu-sheng:
replace by https://github.com/theopenlab/openlab-zuul-jobs/pull/435

PR #443, opened 2019-02-26
Support to enable Ironic on DevStack

iurygregory:
This PR adds the support to enable Ironic on DevStack.

To have ironic enable you need to set `ironic` in `enable_services`

Closes: theopenlab/openlab#195

### DIFF
--- a/roles/create-devstack-local-conf/tasks/main.yml
+++ b/roles/create-devstack-local-conf/tasks/main.yml
@@ -265,3 +265,38 @@
     executable: /bin/bash
   when:
     - '"ceph" in enable_services'
+
+- name: create devstack local conf with ironic enabled
+  shell:
+    cmd: |
+      set -e
+      set -x
+      cat << EOF >> /tmp/dg-local.conf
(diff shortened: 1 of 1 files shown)


roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -265,3 +265,38 @@
...
+  shell:
+    cmd: |
+      set -e
+      set -x
+      cat << EOF >> /tmp/dg-local.conf
+      enable_plugin ironic https://git.openstack.org/openstack/ironic
+      DEFAULT_INSTANCE_TYPE: baremetal
+      OVERRIDE_PUBLIC_BRIDGE_MTU: 1400
+      VIRT_DRIVER=ironic
+      BUILD_TIMEOUT=720
```
juliakreger: Note: We only use 720 seconds in OpenStack CI due to noisy neighbors. 600 should be fine in most cases.

iurygregory: Thanks @juliakreger I think we can go with this and see how it will work on gophercloud so we can tune the values.

PR #450, opened 2019-02-26
Add ARM64 nodesets

mrhillsman:
VEXXHOST only

1. ubuntu-xenial-vxh-arm64
2. ubuntu-bionic-vxh-arm64
3. centos-7-vxh-arm64

Fixes theopenlab/openlab#198
Related-Bug: theopenlab/openlab#194

### DIFF
--- a/zuul.d/nodesets.yaml
+++ b/zuul.d/nodesets.yaml
@@ -25,10 +25,22 @@
         label: ubuntu-xenial-vexxhost
 
 - nodeset:
+    name: ubuntu-xenial-vxh-arm64
+    nodes:
+      - name: ubuntu-xenial-vxh-arm64
+        label: ubuntu-xenial-vxh-arm64
+
+- nodeset:
     name: ubuntu-bionic-vxh-arm64
     nodes:
       - name: ubuntu-bionic-vxh-arm64
         label: ubuntu-bionic-vxh-arm64
+
+- nodeset:
+    name: centos-7-vxh-arm64
+    nodes:
+      - name: centos-7-vxh-arm64
+        label: centos-7-vxh-arm64
 
 - nodeset:
     name: ubuntu-xenial-huaweicloud


liu-sheng:
Hi Melvin, as we have reached an agreement, we should open all these issues in `https://github.com/theopenlab/openlab/issues`

kiwik:
Like @liu-sheng say, please ensure all of these nodesets have been availiable in nodepool, then we can merge the PR.
And I think we should remove `vxh` in nodeset name , s/centos-7-vxh-arm64/centos-7-arm64/ , might other cloud vendor provide ARM resource for the label in the future, if it's not necessary, we should not associate with specified vendor.

kiwik:
@mrhillsman please go on your PR, thanks https://github.com/theopenlab/openlab/issues/198#issuecomment-473180894

zuul.d/nodesets.yaml
```diff
@@ -24,12 +24,24 @@
       - name: ubuntu-xenial-vexxhost
         label: ubuntu-xenial-vexxhost
 
+- nodeset:
+    name: ubuntu-xenial-vxh-arm64
```
kiwik: Only `ubuntu-xenial-arm64` is available now in OpenLab, please update nodeset name and remove the following two ARM nodeset, thx.

Yikun:
Do we still these nodesets.yaml? I prefer we could add this nodesets setting when we really need it. Otherwise, we should maintenance many redundant settings.

kiwik:
> Do we still need these in nodesets.yaml? I prefer we could add this nodesets setting when we really need it. Otherwise, we should maintenance many redundant settings.

We hope to keep consist between Nodepool and Zuul, so reasonable steps is：
1. Config, test and enable labels in Nodepool side
2. Add available labels in openlab-zuul-jobs side

PR #451, opened 2019-02-28
Fix trove job errors

h00130372:
Fix trove job errors with several changes:
1. trovestack scripts can not run as root
2. change path of devstack src to /opt/stack/new/devstack
3. change dest path of trove to /opt/stack/new/trove
4. change owner of files in directory of /opt/stack
   and /etc/trove to support running tox -etrovestack
5. there is an error of trove upstream when exec tox -etrovestack,
   which hard code of /opt/stack/trove/test-upper-constraints.txt
   in tox.ini, change it to
   {toxinidir}/stack/trove/test-upper-constraints.txt

Closes: theopenlab/openlab#183

### DIFF
--- a/playbooks/terraform-provider-openstack-acceptance-test-trove/run.yaml
+++ b/playbooks/terraform-provider-openstack-acceptance-test-trove/run.yaml
@@ -6,12 +6,16 @@
       enable_services:
         - 'trove'
     - install-devstack
+
+- hosts: all
   tasks:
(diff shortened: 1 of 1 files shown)


liu-sheng:
this should still need debuging, I prefer we'd better to mark PR `WIP` if still need to testing. to be standardize the process, we'd better make sure the jobs ok before merging PR.

PR #472, opened 2019-03-20
Support kubernetes version for minikube start

h00130372:
Add a parameter "kubernetes_version" for
role create-single-k8s-cluster-with-minikube to
support specify k8s version if needed.

Related: theopenlab/openlab#213

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -1089,23 +1089,28 @@
       - orange_credentials
 
 - job:
-    name: spark-integration-test-kubeadm-k8s
+    name: spark-integration-test-kubeadm-k8s-v1.14.0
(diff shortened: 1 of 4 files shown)


h00130372:
recheck

zuul.d/jobs.yaml
```diff
@@ -1089,20 +1089,24 @@
       - orange_credentials
 
 - job:
-    name: spark-integration-test-kubeadm-k8s
+    name: spark-integration-test-kubeadm-k8s-latest
```
kiwik: please pin to release `v1.13.4`

h00130372: will pin to the latest release v1.14.0

PR #473, opened 2019-03-21
Update jobs naming notations

kiwik:
- Add version option
- Add kubernetes backend example

Related-Bug: theopenlab/openlab#188


kiwik:
> just remembered that we have discussed do not add version num in jobs name. and I have changed all the jobs with version number in name..

@liu-sheng The `version` is backend version, like: OpenStack Mitaka, Pike, Ocata and Kubernetes 1.10, 1.11, 1.12, it's not project version

PR #478, opened 2019-03-22
Check for branch master before docker push

mrhillsman:
Until a better solution is determined we can check
for the zuul.branch variable to determine whether
we should push images to docker. This will resolve
https://github.com/theopenlab/openlab/issues/223

### DIFF
--- a/playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance/run.yaml
+++ b/playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance/run.yaml
@@ -127,36 +127,28 @@
     - name: Run image build and publish for cloud-provider-openstack
       shell:
         cmd: |
-          # NOTE(flaper87): Export these variables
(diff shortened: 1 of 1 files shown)


playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance/run.yaml
```diff
@@ -127,37 +127,29 @@
...
-          export VERSION=latest
-
-          set -x
-          set -e
-          set -o pipefail
-
-          if [ '{{ zuul.pipeline }}' != 'periodic' ]; then
-              exit 0;
+          if [ '{{ zuul.pipeline }}' == 'periodic' ] && [ '{{ zuul.branch }}' == 'master' ]; then
+            export LOG_DIR='{{ k8s_log_dir }}'
```
mrhillsman: Adding this because export does not make variables available across playbooks when executable is used; it creates a new shell session.

PR #482, opened 2019-03-29
Update KIND to new pipeline format and add cloud-provider-openstack 1.14

mrhillsman:


liu-sheng:
+1 for moving the periodic jobs into new periodic pipelines, that make we can make the load more effective. sorry, I was missing discussion before. Keivin is working on re factoring the galaxy role to make things more simplier ASAP.

zuul.d/projects.yaml
```diff
@@ -37,10 +37,10 @@
             branches: master
         - spark-integration-test-minikube-k8s:
             branches: master
-            
+
 - project:
     name: kubernetes-sigs/kind
-    periodic:
+    periodic-0/12:
```
kiwik: How about putting kind periodic jobs into `periodic-2/14` pipeline? there are less jobs in `periodic-2/14`， we hope to split all of periodic jobs into each pipeline as evenly as possible.

https://docs.openlabtesting.org/publications/contributors/development-guide#periodic-job-and-pipeline-schedule

zuul.d/projects.yaml
```diff
@@ -137,7 +137,7 @@
 
 - project:
     name: kubernetes-sigs/kind
-    periodic:
+    periodic-6/18:
```
kiwik: Why kind jobs have to be triggered 4 times in one day? for most of projects, 2 times is enough

PR #483, opened 2019-03-29
Add a job for checking labkeeper deployment(allinone)

liu-sheng:
Related: # theopenlab/openlab#217

### DIFF
--- a/playbooks/labkeeper-allinone-deployment-test/run.yaml
+++ b/playbooks/labkeeper-allinone-deployment-test/run.yaml
@@ -1,0 +1,37 @@
+- hosts: all
+  tasks:
+    - name: install required packages
+      become: true
+      shell: |
+        apt update -y
+        apt install python python-pip python3 python3-pip kpartx qemu-utils curl python-yaml \
+            debootstrap libffi-dev libssl-dev -y
+        pip install ansible
+      args:
+        executable: /bin/bash
+
+    - name: Prepare ubuntu user for running labkeeper
+      become: yes
+      no_log: yes
+      shell: |
+        useradd -m -d /home/ubuntu -s /bin/bash ubuntu
+        echo ubuntu:ubuntu | chpasswd
+        su - ubuntu -c "
+            ssh-keygen -f /home/ubuntu/.ssh/id_rsa -t rsa -N ''
+            cat /home/ubuntu/.ssh/id_rsa.pub > /home/ubuntu/.ssh/authorized_keys
+        "
+      args:
+        executable: /bin/bash
+
+    - name: Run labkeeper functional test
+      shell: |
+        sudo cp /home/ubuntu/.ssh/id_rsa ./openlab.pem
+        sudo chmod 600 openlab.pem
+        sudo chown ubuntu.ubuntu openlab.pem
+        export DPLOY_TYPE=test-allinone
+        export ANSIBLE_HOST_KEY_CHECKING=False
+        ansible-playbook playbooks/site.yaml -i inventory/test-allinone.yaml -e \
+            "zuul_user_name=zuul-test zuul_user_group=zuul-test"
+      args:
+        executable: /bin/bash
+        chdir: '{{ zuul.project.src_dir }}'
--- a/playbooks/labkeeper-allinone-deployment-test/run.yaml
+++ b/playbooks/labkeeper-allinone-deployment-test/run.yaml
@@ -33,7 +33,11 @@
         chmod 600 openlab.pem
         chown ubuntu.ubuntu openlab.pem
         echo 'youneverknow' > vault-password.txt
+<<<<<<< HEAD
         # Ansible will not load ansible.cfg automatically because this directory is world writable
+=======
+        # Ansible will not load ansible.cfg autommatically because this directory is world writable
+>>>>>>> rebase with master
         export ANSIBLE_CONFIG=ansible.cfg
         ansible-playbook playbooks/site.yaml -i inventory/test-allinone.yaml -e \
             "zuul_user_name=zuul-test zuul_user_group=zuul-test"
--- a/playbooks/labkeeper-allinone-deployment-test/run.yaml
+++ b/playbooks/labkeeper-allinone-deployment-test/run.yaml
@@ -33,7 +33,7 @@
         chmod 600 openlab.pem
         chown ubuntu.ubuntu openlab.pem
         echo 'youneverknow' > vault-password.txt
-        # Because ansible will not load ansible.cfg automatically because this directory is world writable
+        # Ansible will not load ansible.cfg automatically because this directory is world writable
         export ANSIBLE_CONFIG=ansible.cfg
         ansible-playbook playbooks/site.yaml -i inventory/test-allinone.yaml -e \
             "zuul_user_name=zuul-test zuul_user_group=zuul-test"
--- a/playbooks/labkeeper-allinone-deployment-test/run.yaml
+++ b/playbooks/labkeeper-allinone-deployment-test/run.yaml
@@ -33,7 +33,7 @@
         chmod 600 openlab.pem
         chown ubuntu.ubuntu openlab.pem
         echo 'youneverknow' > vault-password.txt
-        # Because ansible will not load ansible.cfg autommatically because this directory is world writable
+        # Because ansible will not load ansible.cfg automatically because this directory is world writable
         export ANSIBLE_CONFIG=ansible.cfg
         ansible-playbook playbooks/site.yaml -i inventory/test-allinone.yaml -e \
             "zuul_user_name=zuul-test zuul_user_group=zuul-test"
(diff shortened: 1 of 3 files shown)


zuul.d/jobs.yaml
```diff
@@ -1114,3 +1114,10 @@
       Run integration tests of Kubernetes IN Docker against ARM64 architecture
     run: playbooks/kind-integration-test-arm64/run.yaml
     nodeset: ubuntu-xenial-arm64
+
+- job:
+    name: labkeeper-allinone-deployment-test
+    parent: init-test
+    description: |
+      Tru to deploy allinone openlab CI environment with labkeeper
```
Yikun: typo? Try to

zuul.d/jobs.yaml
```diff
@@ -1114,3 +1114,10 @@
       Run integration tests of Kubernetes IN Docker against ARM64 architecture
     run: playbooks/kind-integration-test-arm64/run.yaml
     nodeset: ubuntu-xenial-arm64
+
+- job:
```
Yikun: Just a question: This job doesn't include the `nodeset` section, does it means this job perhaps run any nodes in env? or run in a default nodes?

I just worry about if it run in a CentOS unfortunately, the `apt install` in job would be invaild.

playbooks/labkeeper-allinone-deployment-test/run.yaml
```diff
@@ -0,0 +1,37 @@
...
+        su - ubuntu -c "
+            ssh-keygen -f /home/ubuntu/.ssh/id_rsa -t rsa -N ''
+            cat /home/ubuntu/.ssh/id_rsa.pub > /home/ubuntu/.ssh/authorized_keys
+        "
+      args:
+        executable: /bin/bash
+
+    - name: Run labkeeper functional test
+      shell: |
+        sudo cp /home/ubuntu/.ssh/id_rsa ./openlab.pem
```
Yikun: Does these sudo is necessary? I guess not.

liu-sheng: labkeeper need the openlab.pem, you can see the `ansible.cfg` in labkeeper source code. hah

zuul.d/jobs.yaml
liu-sheng: now the default nodeset is `ubuntu-xenial`, because all the jobs inherits from the `base` job, which defining the nodeset as `ubuntu-xenial`

zuul.d/jobs.yaml
liu-sheng: yes, let me fix :)

zuul.d/jobs.yaml
Yikun: Cool, found it, thanks

[1] https://github.com/theopenlab/openlab-zuul-jobs/blob/master/zuul.d/jobs.yaml#L22

playbooks/labkeeper-allinone-deployment-test/run.yaml
Yikun: I mean these `sudo` : ). After F2F discussion, we can set
```
become: true
```
Good to learn it!

playbooks/labkeeper-allinone-deployment-test/run.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Run labkeeper functional test
+      shell: |
+        set -ex
+        cp /home/ubuntu/.ssh/id_rsa ./openlab.pem
+        chmod 600 openlab.pem
+        chown ubuntu.ubuntu openlab.pem
+        echo 'youneverknow' > vault-password.txt
+        # Because ansible will not load ansible.cfg autommatically because this directory is world writable
+        export ANSIBLE_CONFIG=ansible.cfg
+        ansible-playbook playbooks/site.yaml -i inventory/test-allinone.yaml -e \
```
wangxiyuan: I notice a file `test-allinone-ha.yaml`, how could it be used?

bzhaoopenstack: For now,  the site.yaml is too complex and necessary for allinone deployment right? If add the allinone ha test job in the future, the test-allinone-ha.yaml is just test with the static test env on the otc?

liu-sheng: @wangxiyuan I have tried to add a job for testing allinone HA mode, but just consider it will use 3 nodes once time....

liu-sheng: @bzhaoopenstack it will not test against on any real cloud provider, just testing the deployment process.

playbooks/labkeeper-allinone-deployment-test/run.yaml
```diff
@@ -0,0 +1,42 @@
...
+        executable: /bin/bash
+
+    - name: Run labkeeper functional test
+      shell: |
+        set -ex
+        cp /home/ubuntu/.ssh/id_rsa ./openlab.pem
+        chmod 600 openlab.pem
+        chown ubuntu.ubuntu openlab.pem
+        echo 'youneverknow' > vault-password.txt
+        # Because ansible will not load ansible.cfg autommatically because this directory is world writable
```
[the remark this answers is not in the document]
liu-sheng: thanks, will correct it

playbooks/labkeeper-allinone-deployment-test/run.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - name: install required packages
+      shell: |
+        apt update -y
+        apt install python python-pip python3 python3-pip kpartx qemu-utils curl python-yaml \
+            debootstrap libffi-dev libssl-dev -y
+        pip install ansible
+      args:
+        executable: /bin/bash
+
+    - name: Prepare ubuntu user for running labkeeper
```
h00130372: Could we use become_user to running labkeeper?

zuul.d/projects.yaml
```diff
@@ -4,6 +4,12 @@
       jobs:
         - openlab-zuul-jobs-check
 
+- project:
+    name: theopenlab/labkeeper
```
kiwik: Good idea, thank you @liu-sheng 
@Yikun , how about adding a periodic job to execute labchecker at here in separated PR? just checking and reporting, operator can recover later if issue exists

playbooks/labkeeper-allinone-deployment-test/run.yaml
```diff
@@ -0,0 +1,42 @@
...
+        executable: /bin/bash
+
+    - name: Run labkeeper functional test
+      shell: |
+        set -ex
+        cp /home/ubuntu/.ssh/id_rsa ./openlab.pem
+        chmod 600 openlab.pem
+        chown ubuntu.ubuntu openlab.pem
+        echo 'youneverknow' > vault-password.txt
+        # Because ansible will not load ansible.cfg automatically because this directory is world writable
```
kiwik: Typo: Because ... because ?

zuul.d/projects.yaml
Yikun: I was consider add a periodic job to check, but it seems this job it's a little bit special (if the job means creating new vm + adding openstack provider config + doing the remote cloud provider check):
- This job should have the `cloud.yaml` info( for now we need user type `vault-password.txt` and then decrypt the template file to generate the `cloud.yaml`.
- This job can **only** check the cloud provider related testcase, the check item is limited.

But the alternative way is that:
1. Set the nodepool vm as [static pool](https://zuul-ci.org/docs/zuul/admin/nodepool_static.html), named as ` nodepool-allinone`
2. Add a check job, nodesets is set ` nodepool-allinone`, then the job will only scheduled to the real nodepool node, and this job will check the real nodepool or zuul env.
3. the failed check will be reported to the issue page of OpenLab.

If we do this periodic like above, we will use the `cloud.yaml` in real env, and all items would be checked in real env.

playbooks/labkeeper-allinone-deployment-test/run.yaml
liu-sheng: sure, oh, you mean the ubuntu user ? it  need user with ssh key

playbooks/labkeeper-allinone-deployment-test/run.yaml
liu-sheng: will fix :)

liu-sheng:
oh, sorry,seems I have broken this branch... :(

liu-sheng:
now ok

zuul.d/projects.yaml
Yikun: After F2F discussion:
1. We can add the labchecker call after deploy.
2. We can add the periodic check using crontab.
3. We can add the check to ansible role.

kiwik: @Yikun , exactly, thank you

PR #485, opened 2019-04-02
Fix Kind job and related roles

ZhengZhenyu:
Fix Kind job and related roles
Related: theopenlab/openlab#230

### DIFF
--- a/playbooks/kind-integration-test-arm64/run.yaml
+++ b/playbooks/kind-integration-test-arm64/run.yaml
@@ -7,6 +7,7 @@
     - andrewrothstein.bazel
   vars:
     - go_version: 1.12
+    - arch: arm64
     - bazel_ver: '0.23.1'
 
   tasks:
@@ -16,7 +17,7 @@
           go get sigs.k8s.io/kind
           go get k8s.io/kubeadm
           go get k8s.io/kubernetes
-          go get k8s.io/test-infra
+          go get k8s.io/test-infra/kubetest
         executable: /bin/bash
       environment: '{{ global_env }}'
     - name: build KIND against k/k master using bazel
@@ -24,7 +25,6 @@
         cmd: |
           cd $GOPATH/src/sigs.k8s.io/kind && go install .
           kind build base-image --image kindest/base:latest --source $GOPATH/src/sigs.k8s.io/kind/images/base --loglevel debug
-          kind build node-image --type bazel --image kindest/node:latest --base-image kindest/base:latest --kube-root=$GOPATH/src/k8s.io/kubernetes/ --loglevel debug
         executable: /bin/bash
       environment: '{{ global_env }}'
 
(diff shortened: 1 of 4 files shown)


playbooks/kind-integration-test-arm64/run.yaml
```diff
@@ -16,15 +17,14 @@
...
+          go get k8s.io/test-infra/kubetest
         executable: /bin/bash
       environment: '{{ global_env }}'
     - name: build KIND against k/k master using bazel
       shell:
         cmd: |
           cd $GOPATH/src/sigs.k8s.io/kind && go install .
           kind build base-image --image kindest/base:latest --source $GOPATH/src/sigs.k8s.io/kind/images/base --loglevel debug
-          kind build node-image --type bazel --image kindest/node:latest --base-image kindest/base:latest --kube-root=$GOPATH/src/k8s.io/kubernetes/ --loglevel debug
         executable: /bin/bash
```
ZhengZhenyu: The image will be build again in kubetest command, so let's not build it here since it takes about 20 mins and we can speed up by remove this

roles/install-docker-ce/tasks/main.yml
```diff
@@ -33,25 +33,26 @@
...
-
-      iptables -F
-      # Restart docker.
-      systemctl daemon-reload
-      systemctl restart docker
+      if [[ "{{ arch }}" == "amd64" ]]; then
+        # Setup daemon.
+        cat > /etc/docker/daemon.json <<EOF
+        {
+          "exec-opts": ["native.cgroupdriver={{ docker_vcgroupdriver }}"],
```
kiwik: ARM don't support cgroup option now, so please only disable `"exec-opts": ["native.cgroupdriver={{ docker_vcgroupdriver }}"],` in daemon.json.

roles/install-docker-ce/tasks/main.yml
```diff
@@ -33,18 +33,29 @@
...
-      {
-        "exec-opts": ["native.cgroupdriver={{ docker_vcgroupdriver }}"],
-        "log-driver": "json-file",
-        "log-opts": {
-          "max-size": "100m"
-        },
-        "storage-driver": "overlay2"
-      }
-      EOF
+      if [[ "{{ arch }}" == "arm64" ]]; then
```
Yikun:
```
cat > /etc/docker/daemon.json <<EOF
{
  "log-driver": "json-file",
  "log-opts": {
    "max-size": "100m"
  },
  "storage-driver": "overlay2"
}
EOF
if [[ "{{ arch }}" == "arm64" ]]; then
  # Add the exec-opts in 1st line
  sed -i '1a \ \ "exec-opts": ["native.cgroupdriver={{ docker_vcgroupdriver }}"],' /etc/docker/daemon.json
fi
```

I guess you want something like above

roles/install-docker-ce/defaults/main.yaml
```diff
@@ -1,3 +1,4 @@
 ---
 docker_ce_version: "latest"
 docker_vcgroupdriver: "systemd"
+arch: amd64
```
mrhillsman: why are we adding amd64 here? arch should be arm?

PR #489, opened 2019-04-03
Update bazel related tasks

mrhillsman:
Add config for aarch64 and tasks to install
bazel based on dist vs installer (compile)


Yikun:
The latest bazel has been added in nodeset `ubuntu-xenial-arm64` by:

```
sudo apt-get update
sudo apt-get install build-essential openjdk-8-jdk python zip unzip
cd ~
mkdir download
cd download/
wget https://github.com/bazelbuild/bazel/releases/download/0.24.1/bazel-0.24.1-dist.zip
unzip bazel-0.24.1-dist.zip
env EXTRA_BAZEL_ARGS="--host_javabase=@local_jdk//:jdk" bash ./compile.sh
cp output/bazel /usr/local/bin
```

REF: https://docs.bazel.build/versions/master/install-compile-source.html#bootstrap-unix

Current we build it (ubuntu with bazel) as a cloud image, so we can directly using nodeset `ubuntu-xenial-arm64` to enable a job in the **ubuntu with bazel** environment.

kiwik:
Yes, like @Yikun say, we have built `bazel` into OpenLab ARM cloud image and it's available to compile `kind` image directly, so we don't need to update `bazel` role any more.

kiwik:
Hi @mrhillsman you still work on this PR? if not, I hope to close this one.

mrhillsman:
No, you mentioned bazel is in image now so closing, thx

PR #491, opened 2019-04-04
Make config-golang work with arm64

ZhengZhenyu:
Make golang work with arm64
Related-Bug: theopenlab/openlab#230


playbooks/kind-integration-test-arm64/run.yaml
```diff
@@ -3,7 +3,8 @@
   become: yes
   roles:
     - role: config-golang
-      go_version: 1.12
+      go_version: 1.12.1
```
liu-sheng: don't need to config  go_version, as it is same as the default value

PR #493, opened 2019-04-11
[Issue #196]Enable GPU AI job

bzhaoopenstack:
This will ignore the CUDA and NVIDIA driver installation when job use
huaweicloud. Now, GPU resources only support Tesla V100 GPU on
huaweicloud with a specific flavor.

Issue link: https://github.com/theopenlab/openlab/issues/196

### DIFF
--- a/roles/create-ai-environment/tasks/main.yaml
+++ b/roles/create-ai-environment/tasks/main.yaml
@@ -132,35 +132,6 @@
 
     executable: /bin/bash
 
-- name: Prepare GPU env (GPU driver and CUDA)
-  when: is_gpu_enabled and nodepool.cloud != 'huaweicloud'
-  shell:
-    cmd: |
-      set -xe
-      apt-key adv --fetch-keys http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/7fa2af80.pub
-      wget http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/cuda-repo-ubuntu1604_9.1.85-1_amd64.deb
-      cd /var/
-      apt purge -y cuda*
-      apt-get remove -y --purge nvidia-418 nvidia-modprobe nvidia-settings
-      cd -
-      apt install -y ./cuda-repo-ubuntu1604_9.1.85-1_amd64.deb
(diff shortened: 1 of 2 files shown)


roles/create-ai-environment/tasks/main.yaml
```diff
@@ -133,7 +133,7 @@
     executable: /bin/bash
 
 - name: Prepare GPU env (GPU driver and CUDA)
-  when: is_gpu_enabled
+  when: is_gpu_enabled and nodepool.cloud != 'huaweicloud'
```
liu-sheng: the `nodepoo.cloud` I think is depend on the cloud provider's name of nodepool's config `nodepool.yaml`. maybe it is not a good idea, if we change the cloud provider name in `nodepool.yaml`, not sure if there is another way...

bzhaoopenstack: It's from nodepool.yaml .  The real source is from clouds.yaml . As we only support GPU resource on huaweicloud, and different flavor means different GPU driver installation. The zuul job can not know what GPU application in backend, and don't care what GPU used. So we decide to use cloudimage for diferent GPU version. Also we hide the real image name like "something-GPU-version-image" to just show a same resource "ubuntu-xe-gpu", only know the provider or cloud from nodepool.yaml configuration. And provider seems easier change than cloud(name), so forbiding GPU resource to run this task right now. Also notes, if there is a cloud provider who give us gpu resources with a same GPU or a single flavor, we can reuse this task.

kiwik: We have multiple cloud providers and multiple GPU types in each cloud provider (M * N), so there are too many details about GPU card type, GPU driver, linux-headers version and NVIDIA driver version. so I prefer to remove this whole block `Prepare GPU env (GPU driver and CUDA)`, and use pre-build cloud images in each cloud provider to hide this details for AI/DL users, they only need to config label `ubuntu-xenial-gpu` to get GPU nodes from nodepool.

liu-sheng: I mean the cloud provider name maybe changed manually sometimes, for an example, we may add a `huaweicloud` provider in another region, and then we may change the cloud provider name to `huaweicloud-bj` and `huaweicloud-hk`. not sure if there is better way, let see others opinions :)

bzhaoopenstack: @kiwik @liu-sheng Thanks for suggestion. Let's using a simple way to fit. I will remove this task. ;-)

PR #496, opened 2019-04-12
Cleanup Bazel related roles

ZhengZhenyu:
Bazel is now packed in base images for CNCF project tests now, and the tests are successfully running, we can get rid of the related role for installing it to make the repo clean and easier to maintain. If the current way of doing the tests is not good enough, we can come up with other ways like build bazel in disk-image-builder or try to get an official bazel source in Ubuntu deb to make the whole process simpler.


kiwik:
These roles can not work in ARM64 platform, so we should remove them from OpenLab repo, and we have built Bazel into cloud image as an instead, all of related works have been completed in https://github.com/theopenlab/openlab/issues/230 , and we plan to create an element of diskimage-build to make the building process of Bazel on ARM transparent and open source.

PR #498, opened 2019-04-15
Fix inference page CORS error

wangxiyuan:
The inference page now is hosted on log server. Set the model url
to be `logs` so that it can follow http CORS rule


Yikun:
LGTM

PR #500, opened 2019-04-15
Refactor for logs and results arch directory

Yikun:
Now we have multiple different custom logs path exist in different OpenLab jobs.

This patch try to build a consist mechanism and usage in order to avoiding end user and developer's confusion:

1. Add $LOGS_PATH, $RESULTS_PATH global env.
2. Prepare the {{ ansible_user_dir }}/workspace/logs, and {{ ansible_user_dir }}/workspace/test_results.

All logs files (like debug log) should be stored in $LOGS_PATH, and the final test_results (like binaries, artifact) should be stored in $RESULTS_PATH.

Close: theopenlab/openlab#238


Yikun:
We have to import these two args into ``jobs.yaml.init-test`` for cross playbook usage, the test can be found in https://github.com/Yikun/arm-openlab-test/pull/6/files

PR #502, opened 2019-04-16
Support backends OpenStack Stein

h00130372:
Support acceptance test against OpenStack Stein release in OpenLab

Add periodic jobs for ansible/packer/docker-machine/manageiq
against OpenStack Stein release.

Related: theopenlab/openlab#231

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -61,6 +61,15 @@
     run: playbooks/terraform-provider-openstack-acceptance-test/run.yaml
 
 - job:
+    name: terraform-provider-openstack-acceptance-test-stein
+    parent: terraform-provider-openstack-acceptance-test
+    description: |
+      Run terraform provider openstack acceptance test on stein branch
+    vars:
+      global_env:
+        OS_BRANCH: stable/stein
+
+- job:
     name: terraform-provider-openstack-acceptance-test-rocky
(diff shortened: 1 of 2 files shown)


kiwik:
Hi @h00130372 , you are trying to do this work https://github.com/theopenlab/openlab/issues/231 , right? If it is, please add link into commit message and PR describe to make others clear.

zuul.d/jobs.yaml
```diff
@@ -454,6 +463,20 @@
     vars:
       os_sdk: openstacksdk
 
+- job:
+    name: openstacksdk-ansible-functional-devstack-stein
+    parent: init-test
+    description: |
+      Run openstacksdk ansible functional tests against a stein devstack using git stable-2.7 branch version of ansible.
```
liu-sheng: please put this description into two lines.

h00130372: I corrected this last time when I found the description only show the first line content on the job web page(http://status.openlabtesting.org/jobs), like the base job https://github.com/theopenlab/openlab-zuul-jobs/blob/master/zuul.d/jobs.yaml#L7, please check this, or I can open an issue to trace this bug.

liu-sheng: we can land this firstly, maybe there is a fix in zuul in future

PR #503, opened 2019-04-17
Add periodic job for envoy and containerd

Yikun:
This patch add the periodic jobs for envoy and containerd arm
build.

- containerd-build-arm64
- envoy-build-arm64

The links [1] [2] was adding the trigger jobs for each pr of envoy and containerd, and this patch is adding
the periodic job in openlab.

[1] https://github.com/theopenlab/envoy/pull/1
[2] https://github.com/theopenlab/containerd/pull/1


Yikun:
We should add containerd/containerd and envoyproxy/envoy to /etc/zuul/main.yaml, and then restart zuul. but now, there still are many running jobs, so I will recheck after these jobs complete and zuul restart.

playbooks/containerd-build/run.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - role: config-golang
+      arch: arm64
+  tasks:
+    - name: Build containerd
+      shell:
+        cmd: |
+          set -xe
+          apt-get update
+          apt-get install -y btrfs-tools libseccomp-dev git pkg-config
+          echo $PATH
```
liu-sheng: do not need these `echo ...` statements. as we have `set -x`, all the scripts running trace will be print out

playbooks/containerd-build/run.yaml
```diff
@@ -0,0 +1,29 @@
...
+          echo $GOPATH
+
+          if [ ! -d "$GOPATH/src/github.com/containerd" ] ; then
```
liu-sheng: if these two jobs are only run in `periodic` pipelines, we don't need the forks for this project.

zuul.d/jobs.yaml
```diff
@@ -1174,3 +1174,19 @@
       - moo_github_account
     vars:
       is_gpu_enabled: true
+
+- job:
+    name: containerd-build-arm64-periodic
```
liu-sheng: it is better do not add the `periodic` field in jobs' name, there is a doc described about how to name jobs. 
https://docs.openlabtesting.org/publications/contributors/development-guide#job-naming-notations

playbooks/containerd-build/run.yaml
Yikun: The forks project(theopenlab/containerd|envoy) is the demo to show how to trigger the job in pr.  : )

zuul.d/jobs.yaml
kiwik: +1

playbooks/containerd-build/run.yaml
kiwik: +1

zuul.d/projects.yaml
```diff
@@ -154,6 +154,20 @@
...
+      jobs:
+        - containerd-build-arm64-periodic:
+            branches: master
+
+- project:
+    name: envoyproxy/envoy
+    periodic-6/18:
+      jobs:
+        - envoy-build-arm64-periodic:
+            branches: master
```
kiwik: As an example, I think we should remove the jobs define and playbooks in OpenLab side, reasonable step should be:
1. Merge job define and playbook in project side (forked containerd and envoy, as an example, official projects in the future) to **enable PR triggered job**
2. Just refer the merged job name in periodic pipeline of OpenLab side to **enable periodic job**

playbooks/containerd-build/run.yaml
Yikun: done

zuul.d/jobs.yaml
Yikun: done

zuul.d/projects.yaml
kiwik: LGTM

kiwik:
Let us to merge this PR after https://github.com/theopenlab/envoy/pull/1 and https://github.com/theopenlab/containerd/pull/1/ are merged.

zuul.d/projects.yaml
```diff
@@ -158,6 +158,20 @@
...
+    periodic-6/18:
+      jobs:
+        - containerd-build-arm64:
+            branches: master
+
+- project:
+    name: theopenlab/envoy
+    periodic-6/18:
+      jobs:
+        - envoy-build-arm64:
```
kiwik: Please split these ARM job into different periodic pipeline, we don't have many ARM node resources, so trigger these jobs in different time to avoid waiting.

Yikun: done

Yikun:
The job is not loaded, we should restart zuul in production env, but there still are many running jobs.

zuul.d/projects.yaml
kiwik: `kind-integration-test-arm64` and `envoy-build-arm64` still are in one periodic pipeline `periodic-6/18`

Yikun: oops, sorry for miss that, will do a update soon.
```
containerd-build-arm64: 2:00/14:00 about 10 mins
envoy-build-arm64: 4:00/16:00 about 1 hour
kind-integration-test-arm64: 06:00/18:00 about 3hours
```

PR #510, opened 2019-05-06
Use ruby 2.4.0 instead for manageiq jobs

huangtianhua:
An error occurred while installing dry-validation (0.13.1),
and Bundler cannot continue for manageiq jobs.
Change to use ruby 2.4.0 to fix this bug.

Related-Bug: theopenlab/openlab#246

### DIFF
--- a/playbooks/manageiq-providers-openstack-test-devstack/run.yaml
+++ b/playbooks/manageiq-providers-openstack-test-devstack/run.yaml
@@ -4,7 +4,8 @@
     #NOTES: manageiq-providers-openstack don't support OpenStack version discovery, so enable legacy endpoint format.
     DISABLE_HTTPD_MOD_WSGI: true
   roles:
-    - config-ruby
+    - role: config-ruby
+      ruby_version: '2.4.0'
     - clone-devstack-gate-to-workspace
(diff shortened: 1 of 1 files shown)


liu-sheng:
tested OK in allinone env ?

liu-sheng:
@theopenlab-ci seems this job cannot pass due to test cases failure,  you can check the previous logs to confirm.

h00130372:
@liu-sheng What do you mean the 'test cases failure'? The earliest log I can found  is http://status.openlabtesting.org/build/f3e0feb399aa4a8789aea2019ed7a114, and the build failed is because the libssl-dev package is not found(this fixed in https://github.com/theopenlab/openlab/issues/242)

huangtianhua:
@liu-sheng @Yikun let's merge this first, the manageiq jobs failed and have several problems, this is the first step to continue run the job.

huangtianhua:
@Yikun Thanks. We disabled the jobs of manageiq against with public clouds almost two months ago, so I think we can change it once the jobs enabled again?

Yikun:
@huangtianhua Yes, I think we can.

PR #515, opened 2019-05-14
workaround to fix cpo e2e job

ramineni:
As per the issue
https://github.com/kubernetes/kubernetes/issues/77521#issuecomment-490108125,
issue seems to be happen only when KUBE_FASTBUILD=true.
Changing the same to check if it fixes the e2e job.
https://github.com/theopenlab/openlab/issues/254


ramineni:
cc @dims @adisky @kiwik

PR #522, opened 2019-05-16
Change the volcano docker image and fix some nit

liu-sheng:
- Change the docker image to make volcano record job start time and end
  time
- Change the timeout of job because it now do not need that long time any more
- Use volcano 0.1 release than master code

### DIFF
--- a/playbooks/tensorflow-volcano-kubeflow-integration-test-k8s/volcano_tfbenchmarks.yaml
+++ b/playbooks/tensorflow-volcano-kubeflow-integration-test-k8s/volcano_tfbenchmarks.yaml
@@ -71,7 +71,7 @@
         set -o pipefail
         set -ex
         mkdir -p src/volcano.sh/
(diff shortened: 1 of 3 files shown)


playbooks/tensorflow-volcano-kubeflow-integration-test-k8s/volcano_tfbenchmarks.yaml
```diff
@@ -71,7 +71,7 @@
         set -o pipefail
         set -ex
         mkdir -p src/volcano.sh/
-        git clone https://github.com/volcano-sh/volcano src/volcano.sh/volcano
+        git clone https://github.com/volcano-sh/volcano src/volcano.sh/volcano -b release-0.1
```
h00130372: Consider the extensibility, use a variable to set the version of volcano?

liu-sheng: sure, will update

PR #538, opened 2019-05-30
Correct the openstack git repo and openstack-infra

Yikun:
1. The OpenStack official repo has been moved to `opendev.org`,
we should also follow it.
2. openstack-infra has been migrated to openstack repo.

Closes: theopenlab/openlab#284


Yikun:
^ looks like we should chaneg the /etc/zuul/main.yaml first.

PR #539, opened 2019-05-30
Fix kind base image entrypoint access rule

ZhengZhenyu:
Fix kind base image entrypoint access rule
Related-Bug: theopenlab/openlab#269


playbooks/kind-integration-test-arm64/run.yaml
```diff
@@ -31,6 +31,7 @@
           mkdir -p $GOPATH/src/sigs.k8s.io/
           kind_tmp=`ls -d $GOPATH/pkg/mod/sigs.k8s.io/kind*`
           ln -sf $kind_tmp $GOPATH/src/sigs.k8s.io/kind
+          sudo chmod +x $GOPATH/src/sigs.k8s.io/kind/images/base/entrypoint
```
Yikun: We already set the host "become: yes", so we do not need sudo.

PR #543, opened 2019-06-03
Enable service q-lbaasv2 when octavia enabled for mitaka

huangtianhua:
Service q-lbaasv2 is required if enable octavia when
installing devstack of mitaka branch.
https://github.com/openstack/octavia/blob/mitaka-eol/devstack/plugin.sh#L253-L255

Closes: theopenlab/openlab#289

### DIFF
--- a/roles/create-devstack-local-conf/tasks/main.yml
+++ b/roles/create-devstack-local-conf/tasks/main.yml
@@ -190,8 +190,14 @@
     cmd: |
       set -e
       set -x
+      # enable q-lbaasv2 for stable/mitaka branch
+      if [[ "{{ global_env.OS_BRANCH }}" == "stable/mitaka" ]];then
+          services='octavia,o-cw,o-hm,o-hk,o-api,q-lbaasv2'
+      else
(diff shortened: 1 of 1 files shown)


roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -77,6 +77,17 @@
...
+- name: enable q-lbaasv2 with octavia enabled for stable/mitaka branch
+  shell:
+    cmd: |
+      set -e
+      set -x
+      cat << EOF >> /tmp/dg-local.conf
+      enable_service q-lbaasv2
+      EOF
+    executable: /bin/bash
+  when: global_env.OS_BRANCH == 'stable/mitaka' and '"octavia" in enable_services'
```
kiwik: We have a code block to add options when `octavia` is enabled in line 204, I prefer to add some logic in there, something like:
```bash
if [[ "{{ global_env.OS_BRANCH }}" == "stable/mitaka" ]]
    enable_service q-lbaasv2
```

PR #544, opened 2019-06-03
Change log dir

ZhengZhenyu:
Change log dir

### DIFF
--- a/playbooks/kind-integration-test-arm64/run.yaml
+++ b/playbooks/kind-integration-test-arm64/run.yaml
@@ -39,7 +39,7 @@
 
     - name: Create e2e directory for testgrid
       file:
-        path: {{ k8s_log_dir }}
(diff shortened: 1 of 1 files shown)


playbooks/kind-integration-test-arm64/run.yaml
```diff
@@ -53,6 +54,6 @@
             --check-version-skew=false \
             --down \
             --test_args="--ginkgo.focus=\[Conformance\] --ginkgo.skip=\[Serial\]|Alpha|Kubectl|\[(Disruptive|Feature:[^\]]+|Flaky)\] --num-nodes=3" \
-            --timeout=240m
+            --timeout=240m | tee $LOG_DIR/e2e.log
```
Yikun: It seems we should create LOG_DIR first.

ZhengZhenyu:
recheck

PR #552, opened 2019-06-11
Extract install openjdk8 to a role

ZhengZhenyu:
There is some unknown error that causes
spark job exit when installing openjdk8,
extract it to be a role and run it at
the begaining of the job to debug the
root causes.

### DIFF
--- a/playbooks/spark-integration-test-kubeadm-k8s/run.yaml
+++ b/playbooks/spark-integration-test-kubeadm-k8s/run.yaml
@@ -1,6 +1,7 @@
 - hosts: all
   become: yes
   roles:
+    - install-openjdk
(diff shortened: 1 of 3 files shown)


playbooks/spark-integration-test-kubeadm-k8s/run.yaml
```diff
@@ -1,6 +1,7 @@
 - hosts: all
   become: yes
   roles:
+    - install-openjdk-8-jdk
```
kiwik:
```suggestion
    - install-openjdk
```

Please add option to specify the jdk version for this new role.

liu-sheng:
just my 2 cents, I didn't find any meaning of this extracting role..

PR #553, opened 2019-06-11
Fix packer-master-functional-devstack for changing "make bin" to "make dev"

bzhaoopenstack:
This fix the wrong platform and ARCH name from 'uname -a', as we use the ubuntu and X86 image, so we force to make packer job use linux_amd64 packer, and copy it into the right place.

Close: theopenlab/openlab#247

### DIFF
--- a/roles/config-packer/tasks/main.yaml
+++ b/roles/config-packer/tasks/main.yaml
@@ -6,13 +6,7 @@
 - name: Install master of Packer from source
   shell: |
     set -ex
-    make bin
-    # Currently, we use ubuntu amd64 image to test, but the make bin failed to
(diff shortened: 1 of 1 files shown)


bzhaoopenstack:
This PR will make as [DNM] I post another fix in packer. After it merge, let's see whether this issue is fix

roles/config-packer/tasks/main.yaml
```diff
@@ -7,6 +7,12 @@
   shell: |
     set -ex
     make bin
```
kiwik: Let's using `make dev` to try

bzhaoopenstack: done

PR #556, opened 2019-06-14
Add hbase periodic build job

bzhaoopenstack:
This PR add a new hbase periodic job at UTC 10:00

Related-Bug: theopenlab/openlab#300

### DIFF
--- a/playbooks/hbase-arm-build/run.yaml
+++ b/playbooks/hbase-arm-build/run.yaml
@@ -71,19 +71,17 @@
           cd "{{ ansible_user_dir }}/{{ zuul.project.src_dir }}"
           sed -i 's#<protocArtifact>com.google.protobuf:protoc:${external.protobuf.version}:exe:${os.detected.classifier}</protocArtifact>#<protocExecutable>/usr/bin/protoc</protocExecutable>#g' pom.xml
           sed -i 's#<protocArtifact>com.google.protobuf:protoc:${internal.protobuf.version}:exe:${os.detected.classifier}</protocArtifact>#<protocExecutable>/usr/local/bin/protoc</protocExecutable>#g' hbase-protocol-shaded/pom.xml
-
-          mvn package -DskipTests | tee $LOGS_PATH/hbase_build.log
-
+          mvn package -DskipTests | tee "{{ ansible_user_dir }}/workspace/logs/hbase_build.log"
           # Copy jar packages
           for i in `find . -name target`; do
               cp_dir=${i%/*}
               cp_dir=${cp_dir#*.}
               if [ ${i%/*} == '.' ]; then
-                  mkdir -p $RESULTS_PATH/hbase/
-                  cp -rf $i $RESULTS_PATH/hbase/
+                  mkdir -p "{{ ansible_user_dir }}/workspace/test_results/hbase/"
+                  cp -rf $i "{{ ansible_user_dir }}/workspace/test_results/hbase/"
               else
-                  mkdir -p $RESULTS_PATH$cp_dir
-                  cp -rf $i $RESULTS_PATH$cp_dir
+                  mkdir -p "{{ ansible_user_dir }}/workspace/test_results$cp_dir"
+                  cp -rf $i "{{ ansible_user_dir }}/workspace/test_results$cp_dir"
               fi
           done
         executable: /bin/bash
--- a/playbooks/hbase-arm-build/run.yaml
+++ b/playbooks/hbase-arm-build/run.yaml
@@ -5,13 +5,12 @@
     - install-maven
   tasks:
     - name: Build hbase
-      shell:
-        cmd: |
+      shell: |
           set -xe
           apt install protobuf-compiler
           # build begin
           mvn package -DskipTests | tee "{{ ansible_user_dir }}/workspace/logs/hbase_build.log"
-        args:
-          executable: /bin/bash
-          chdir: '{{ zuul.project.src_dir }}'
+      args:
+        executable: /bin/bash
+        chdir: '{{ zuul.project.src_dir }}'
       environment: '{{ global_env }}'
--- a/playbooks/hbase-arm-build/run.yaml
+++ b/playbooks/hbase-arm-build/run.yaml
@@ -7,7 +7,7 @@
     - name: Build hbase
       shell: |
         set -xeo pipefail
-        apt install protobuf-compiler
+        apt install protobuf-compiler -y
         # build begin
         mvn package -DskipTests | tee "{{ ansible_user_dir }}/workspace/logs/hbase_build.log"
       args:
--- a/playbooks/hbase-arm-build/run.yaml
+++ b/playbooks/hbase-arm-build/run.yaml
@@ -4,7 +4,6 @@
     - install-openjdk
     - install-maven
   tasks:
-
     - name: Build hbase
       shell:
         cmd: |
--- a/playbooks/hbase-arm-build/run.yaml
+++ b/playbooks/hbase-arm-build/run.yaml
@@ -6,10 +6,10 @@
   tasks:
     - name: Build hbase
       shell: |
-          set -xe
-          apt install protobuf-compiler
-          # build begin
-          mvn package -DskipTests | tee "{{ ansible_user_dir }}/workspace/logs/hbase_build.log"
+        set -xe
+        apt install protobuf-compiler
+        # build begin
+        mvn package -DskipTests | tee "{{ ansible_user_dir }}/workspace/logs/hbase_build.log"
       args:
         executable: /bin/bash
         chdir: '{{ zuul.project.src_dir }}'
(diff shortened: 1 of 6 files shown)


bzhaoopenstack:
recheck

mrhillsman:
I do not see a request filed for this.

bzhaoopenstack:
Hi @Melvin, 
Thanks for reminding, already post an new issue(Test request) for this. 
https://github.com/theopenlab/openlab/issues/300

playbooks/hbase-arm-build/run.yaml
```diff
@@ -0,0 +1,87 @@
...
+          set -xe
+          # install java
+          apt-get update
+          apt install default-jre -y
+          apt install default-jdk -y
+          export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-arm64
+          export PATH=$PATH:$JAVA_HOME/bin
+          echo "export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-arm64" >> ~/.profile
+          echo "export PATH=$PATH:$JAVA_HOME/bin" >> ~/.profile
+          source ~/.profile
```
kiwik: OpenLab have a role `install-openjdk` to config JDK, you can do refactor and use it.

playbooks/hbase-arm-build/run.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+          # install mvn
+          apt install maven -y
```
kiwik: I suggest to download and config Maven from binary package with specific version, don't use apt, because some projects might depends on specific version, like: Spark, and please add a new role to do it

playbooks/hbase-arm-build/run.yaml
```diff
@@ -0,0 +1,87 @@
...
+          cd "{{ ansible_user_dir }}/{{ zuul.project.src_dir }}"
+          sed -i 's#<protocArtifact>com.google.protobuf:protoc:${external.protobuf.version}:exe:${os.detected.classifier}</protocArtifact>#<protocExecutable>/usr/bin/protoc</protocExecutable>#g' pom.xml
+          sed -i 's#<protocArtifact>com.google.protobuf:protoc:${internal.protobuf.version}:exe:${os.detected.classifier}</protocArtifact>#<protocExecutable>/usr/local/bin/protoc</protocExecutable>#g' hbase-protocol-shaded/pom.xml
+          mvn package -DskipTests | tee "{{ ansible_user_dir }}/workspace/logs/hbase_build.log"
+          # Copy jar packages
+          for i in `find . -name target`; do
+              cp_dir=${i%/*}
+              cp_dir=${cp_dir#*.}
+              if [ ${i%/*} == '.' ]; then
+                  mkdir -p "{{ ansible_user_dir }}/workspace/test_results/hbase/"
```
kiwik: Please don't save jars in OpenLab log server

bzhaoopenstack:
recheck

zuul.d/jobs.yaml
```diff
@@ -2141,3 +2141,12 @@
       - OS:ubuntu-xenial
       - Arch:x86_64
       - BuildType:Integration test
+
+# hbase arm build periodic job
+- job:
+    name: hbase-build-arm64
+    parent: init-test
+    description: |
+      Hbase build in openlab cluster.
```
kiwik:
```suggestion
      Hbase build in openlab arm instance.
```

zuul.d/jobs.yaml
```diff
@@ -2141,3 +2141,12 @@
...
+    run: playbooks/hbase-arm-build/run.yaml
+    nodeset: ubuntu-xenial-arm64
```
kiwik: Please add tags for the new job.

roles/install-maven/tasks/main.yaml
```diff
@@ -0,0 +1,19 @@
...
+- name: Set maven env vars
+  set_fact:
+    maven_env:
+      PATH: '{{ ansible_user_dir }}/{{ zuul.project.src_dir }}/apache-maven-{{ maven_version }}/bin:{{ ansible_env.PATH }}'
+  no_log: yes
+
+- name: Merge maven env vars into global env
+  set_fact:
+    global_env: '{{ global_env | combine(maven_env) }}'
+  no_log: yes
```
kiwik: Please add a task to show `mvn` path and version:
```bash
which mvn
mvn --version
```

roles/install-openjdk/tasks/main.yaml
```diff
@@ -5,3 +5,20 @@
...
+  set_fact:
+    java_env:
+      JAVA_HOME: "{{ real_java_home.stdout }}"
+      PATH: '{{ real_java_home.stdout }}/bin:{{ ansible_env.PATH }}'
+  no_log: yes
+
+- name: Merge java env vars into global env
+  set_fact:
+    global_env: '{{ global_env | combine(java_env) }}'
+  no_log: yes
```
kiwik: Please add a task to show `javac` path and version:
```
which javac
javac -version
```

zuul.d/jobs.yaml
bzhaoopenstack: Done

zuul.d/jobs.yaml
bzhaoopenstack: Done

roles/install-openjdk/tasks/main.yaml
bzhaoopenstack: Done

roles/install-maven/tasks/main.yaml
bzhaoopenstack: Done

roles/install-maven/tasks/main.yaml
```diff
@@ -0,0 +1,28 @@
+- name: Install Apache Maven {{ maven_version }}
+  shell: |
+    wget "https://archive.apache.org/dist/maven/maven-3/{{ maven_version }}/binaries/apache-maven-{{ maven_version }}-bin.tar.gz"
```
wangxiyuan: seems this url is unreachable from some IPs, such as China. Suppose someday we have VMs located in China, this job won't work. How about use another mirror url, or try more than one url.

liu-sheng: maybe we can just assume all the nodes are outside GFW, also there are some other urls such as google.com in `openlab-zuul-jobs`

liu-sheng: hmm, I just checked. the `archive.apache.org` is the old release URL, now it is no longer recommend to use, please use `www.apache.org` instead. that is the reason we cannot open the url  in browser I think, see:
http://www.apache.org/dev/mirrors#location

wangxiyuan: emm, sounds reasonable. My concern is that HuaweiCloud ARM resources are now only provided inside GFW.  If we use them, we'll hit the issue. Leave a TODO here is OK currently.

bzhaoopenstack: Thanks very much for kind suggestion. ;-)

zuul.d/jobs.yaml
```diff
@@ -2139,3 +2139,19 @@
...
+    name: hbase-build-arm64
+    parent: init-test
+    description: |
+      Hbase build in openlab arm instance.
+    run: playbooks/hbase-arm-build/run.yaml
+    nodeset: ubuntu-xenial-arm64
+    tags:
+      - Category:BigData
+      - Project:apache/hbase
+      - Application:Hbase@latest-release
```
kiwik: `latest-release` means released version, `master` means developing branch

```suggestion
      - Application:Hbase@master
```

zuul.d/jobs.yaml
```diff
@@ -2139,3 +2139,19 @@
...
+      - OS:ubuntu-xenial
+      - Arch:aarch_64
```
kiwik: This arch should keep consistent with `uname -i`, and please fix job `kind-integration-test-arm64`' arch option
```suggestion
      - Arch:aarch64
```

bzhaoopenstack: Thanks, Done

zuul.d/jobs.yaml
bzhaoopenstack: Thanks, Done

zuul.d/jobs.yaml
```diff
@@ -2139,3 +2139,19 @@
...
+      Hbase build in openlab arm instance.
+    run: playbooks/hbase-arm-build/run.yaml
+    nodeset: ubuntu-xenial-arm64
+    tags:
+      - Category:BigData
+      - Project:apache/hbase
+      - Application:Hbase@master
+      - OS:ubuntu-xenial
+      - Arch:aarch64
+      - BuildType:Integration test
```
wangxiyuan: Not sure if  we have any other BuildType? Seems this job only does `compile` and  `package` action, no test at all. Does `Integration test` is suitable here? Or you'll add the `test` action later?

kiwik: Good catch, according to @huangtianhua 's suggestion, the following types are options
1. Integration test
2. Unit test
3. Compile

bzhaoopenstack: Thanks. It's now clear.

PR #560, opened 2019-06-18
Debug openstack provider image build

wangxiyuan:
docker login failed, need to login the job environment to debug.


wangxiyuan:
No need, restart zuul service can solve this problem.

PR #568, opened 2019-06-24
Use "go install" instead of "make" command for compiling packer

liu-sheng:
The "make bin" command has been broken(fixed but not released)
The "make dev" command can only work on master branch

Closes: #theopenlab/openlab#306


liu-sheng:
recheck

PR #570, opened 2019-06-25
Make kind job using stable versions

ZhengZhenyu:
This change make the kind on arm job using
stable versions (kind@v0.3.0 + k8s@v1.14.3)
as it is the most other jobs used in test
grids and will be easier for debuging as it
is the firt kind on arm job.

### DIFF
--- a/playbooks/kind-integration-test-arm64/run.yaml
+++ b/playbooks/kind-integration-test-arm64/run.yaml
@@ -14,10 +14,12 @@
     - name: git required repositories
       shell:
         cmd: |
-          GO111MODULE="on" go get sigs.k8s.io/kind@master
+          GO111MODULE="on" go get sigs.k8s.io/kind@v0.3.0
           go get k8s.io/kubeadm
           go get k8s.io/kubernetes
           go get k8s.io/test-infra/kubetest
+          cd $GOPATH/src/k8s.io/kubernetes
+          git checkout v1.14.3
         executable: /bin/bash
       environment: '{{ global_env }}'
     - name: build KIND against k/k master using bazel


ZhengZhenyu:
Kind v0.4.0 just released, I will test to see if 0.4.0 is better.

kiwik:
I think we can use the `kind-integration-test-arm64` as base job, and add kind version option for its child job in order to test multiple versions in the future.

playbooks/kind-integration-test-arm64/run.yaml
```diff
@@ -14,10 +14,12 @@
     - name: git required repositories
       shell:
         cmd: |
-          GO111MODULE="on" go get sigs.k8s.io/kind@master
+          GO111MODULE="on" go get sigs.k8s.io/kind@v0.3.0
```
liu-sheng: need also to update the job's tag: https://github.com/theopenlab/openlab-zuul-jobs/blob/master/zuul.d/jobs.yaml#L2002
and I am not sure, why here need to get kind source code again ? if the project to trigger this job is `kubernetes-sigs/kind`, the source code should already be downloaded, in '{{ zuul.project.src_dir }}'

ZhengZhenyu: This may sounds a little bit weird but for kind with version > v0.3.0, it have to be downloaded with `GO111MODULE="on"` mode to be seccessfully running because of https://github.com/kubernetes-sigs/kind/issues/532, so we have to download it in `module` mode to get all the dependency installed correctly first. And I just found another weird thing is that in the latest v0.4.0, there is a command https://github.com/kubernetes-sigs/kind/blob/master/hack/build/go_container.sh#L22 that uses `git` command, and what I've downloaded by using `go get` in module mode is not a git repo, so after installing all the dependence, I have to actually go back to a git cloned repo to checkout to v0.4.0 to actually run it. Any suggestion?

liu-sheng: I think you can try:
```
cd '{{ zuul.project.src_dir }}'
go get ./...
```
not sure if it can work, but that will also download dependencies

kiwik: According to https://github.com/kubernetes-sigs/kind#installation-and-usage
1. upgrade golang to 1.12.6
2. like @liu-sheng say, zuul have download kind master into '{{ zuul.project.src_dir }}'/sigs.k8s.io/kind , you don't need to go get again, and please try to use `go mod download` to handle the go.mod dependences
3. you can add `k8s.io/kubeadm`, `k8s.io/kubernetes` and `k8s.io/test-infra/kubetest` into job required-projects, that is zuul build-in function to avoid `go get` again
Then you can remove the first two tasks, short the job define

ZhengZhenyu:
Thanks for the comment, I will close this one and start a new one for total refactor of this job

PR #571, opened 2019-06-26
Add new role "install-docker"

kiwik:
The new role "install-docker" install docker-ce by using get.docker.com
script, it support multiple OS and arch, and keep consistent docker-ce
version, mark role "install-docker-ce" as deprecated, and apply new role
in all of docker-machine jobs. Update tag of docker-machine job from
latest-release to 18.06 in order to avoid confusions.

Related-Bug: theopenlab/openlab#308

### DIFF
--- a/roles/install-docker/tasks/main.yml
+++ b/roles/install-docker/tasks/main.yml
@@ -1,0 +1,34 @@
+- name: Install Docker CE {{ docker_version }} using get.docker.com
+  shell:
+    cmd: |
+      set -x
+      set -e
+
+      # Download docker install script
+      # The script worked on multiple OS (Ubuntu, CentOS, SUSE) and Archs (x86_64 and aarch64)
+      # and have consistent docker-ce version, like: 18.06, 18.09
+      curl -fsSL https://get.docker.com -o get-docker.sh
+
+      # Install docker ce.
+      if [[ "{{ docker_version }}" == "latest" ]]; then
+          sh get-docker.sh
+      else
+          VERSION="{{ docker_version }}" sh get-docker.sh
+      fi
+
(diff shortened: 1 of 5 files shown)


roles/install-docker/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
+- name: Install Docker CE {{ docker_version }} using get.docker.com
+  shell:
+    cmd: |
+      set -x
+      set -e
+
+      # Download docker install script
+      # The script worked on multiple OS (Ubuntu, CentOS, SUSE) and Archs (x86_64 and aarch64)
+      # and have consistent docker-ce version, like: 18.06, 18.09
+      curl -fsSL https://get.docker.com -o get-docker.sh
```
liu-sheng: not sure if there "docker_vcgroupdriver" setting is required.
https://github.com/theopenlab/openlab-zuul-jobs/blob/master/playbooks/spark-integration-test-minikube-k8s/run.yaml#L6

kiwik: I check docker document, default value of this option is `cgroupfs`, and kubelet also use `cgroupfs` as file system. I don't why role `install-docker-ce` use `systemd` as default value, and specify `cgroupfs` in the job, so I think we can remove it to avoid more confusion.

liu-sheng: make sence, let's use this new role and see if it break anything

huangtianhua: install-docker-ce use systemd according the doc https://kubernetes.io/docs/setup/production-environment/container-runtimes/#docker, but I think it should be cgroupfs :)

PR #572, opened 2019-06-26
Add job support run Apache Hive building on ARM64 server

liu-sheng:
Implement: theopenlab/openlab#303

### DIFF
--- a/playbooks/hive-build-arm64/run.yaml
+++ b/playbooks/hive-build-arm64/run.yaml
@@ -1,0 +1,14 @@
+- hosts: all
+  become: yes
+  roles:
+    - install-openjdk
+    - install-maven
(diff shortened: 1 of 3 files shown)


playbooks/hive-build-arm64/run.yaml
```diff
@@ -0,0 +1,14 @@
+- hosts: all
+  become: yes
+  roles:
+    - install-openjdk
+    - install-maven
+  tasks:
+    - name: install required packages
```
bzhaoopenstack: This task name here should be "Build Hive" something.

liu-sheng: yes, my mistake, thanks for pointing out.

PR #575, opened 2019-06-28
Apply role "install-docker" to all of jobs

kiwik:
Make all of OpenLab jobs to use new role "install-docker" to
install and config docker-ce, update tags of related jobs
to specific docker version and remove all of "latest-release" define.

Closes: theopenlab/openlab#308


ZhengZhenyu:
why not get rid of install-docker-ce role in this patch directly?

kiwik:
@ZhengZhenyu will remove role `install-docker-ce` when all of related job builds are OK

PR #576, opened 2019-06-28
Refactor KinD on ARM job

ZhengZhenyu:
Refactor KinD on ARM job

### DIFF
--- a/playbooks/kind-integration-test-arm64/run.yaml
+++ b/playbooks/kind-integration-test-arm64/run.yaml
@@ -3,66 +3,45 @@
   become: yes
   roles:
     - role: config-golang
-      go_version: 1.12.1
+      go_version: 1.12.6
       arch: arm64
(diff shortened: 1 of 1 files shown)


playbooks/kind-integration-test-arm64/run.yaml
```diff
@@ -3,66 +3,45 @@
   become: yes
   roles:
     - role: config-golang
-      go_version: 1.12.1
+      go_version: 1.12.6
       arch: arm64
     - role: install-docker-ce
```
liu-sheng: this will be removed, and not sure if the new `install-docker` role can support ARM ?
https://github.com/theopenlab/openlab-zuul-jobs/pull/575

ZhengZhenyu: it supports, but since that one didn't merged, I still write this way

PR #579, opened 2019-06-29
Add a no comments pipeline

Yikun:
This patch adds a new pipeline, this pipeline do not comment
when the ci set results.

An related patch is that the result detail will set to
https://status.openlabtesting.org/builds?project={change.project}&change={change.number},{change.patchset}&pipeline={pipeline.name}
which would be a separated patch in zuul.conf. See https://github.com/theopenlab/labkeeper/pull/163

close: https://github.com/theopenlab/openlab/issues/311


zuul.d/pipelines.yaml
```diff
@@ -30,6 +30,38 @@
...
+          comment: (?i)^\s*recheck\s*$
+    start:
+      github:
+        status: pending
+        comment: false
+    success:
+      github:
+        status: 'success'
+        comment: false
+        status_url=https://status.openlabtesting.org/builds?project={change.project}&change={change.number},{change.patchset}&pipeline={pipeline.name}
```
wangxiyuan: status-url

Yikun: Yes, I move it to zuul.conf

https://zuul-ci.org/docs/zuul/admin/drivers/github.html#attr-pipeline.%3Creporter%3E.%3Cgithub%20source%3E.status-url

PR #580, opened 2019-07-01
Fix install-docker role

kiwik:
1. systemctl is not available on Ubuntu 14.04, so using
   "service docker restart" as a workaround.
2. change option "--r" position in spark-inegration-test-minikube-k8s
   job to avoid "--r" is bypassed to mvn.
3. change the importing order of roles in docker-machine job to make sure the
  docker installation won't clean the iptables rules added by devstack

Related-Bug: theopenlab/openlab#308
Related-Bug: theopenlab/openlab#310

Co-authored-by: liu-sheng <liusheng2048@gmail.com>

### DIFF
--- a/playbooks/docker-machine-functional-devstack/run.yaml
+++ b/playbooks/docker-machine-functional-devstack/run.yaml
@@ -1,10 +1,10 @@
 - hosts: all
   become: yes
   roles:
+    - config-docker-machine
     - clone-devstack-gate-to-workspace
     - create-devstack-local-conf
     - install-devstack
-    - config-docker-machine
   tasks:
     - name: Run Integration tests of Docker machine against devstack
       shell:
--- a/roles/install-docker/tasks/main.yml
+++ b/roles/install-docker/tasks/main.yml
@@ -15,4 +15,20 @@
       else
           VERSION="{{ docker_version }}" sh get-docker.sh
       fi
+
+      # Stopping firewall and allow all traffic
+      iptables -F
+      iptables -X
+      iptables -t nat -F
+      iptables -t nat -X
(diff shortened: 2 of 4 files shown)


roles/install-docker/tasks/main.yml
```diff
@@ -27,8 +27,14 @@
       iptables -P FORWARD ACCEPT
       iptables -P OUTPUT ACCEPT
 
-      # Restart docker.
-      systemctl daemon-reload
-      systemctl restart docker
+      # NOTE: systemctl is not available in Ubuntu 14.04
```
liu-sheng: maybe just use `service docker restart` ?

kiwik: OK, good point.

liu-sheng:
sorry, according to my testing, these jobs still need docker installed execution node.

playbooks/docker-machine-functional-devstack/run.yaml
```diff
@@ -1,10 +1,10 @@
 - hosts: all
   become: yes
   roles:
+    - config-docker-machine
     - clone-devstack-gate-to-workspace
     - create-devstack-local-conf
     - install-devstack
-    - config-docker-machine
```
bzhaoopenstack: Just want to make sure, does it pass in allinone? Others looks good.

liu-sheng: yes, you can see: http://80.158.17.129/builds?uuid=f62c7423653d41a2a9cff9777cb4ce3a

kiwik:
LGTM, but I'am PR owner, so can't approve

PR #581, opened 2019-07-04
Set start detail url to status main page

Yikun:
In this patch, we change the pipeline start section, that means when start to running job, the status url will be set to `https://status.openlabtesting.org/status` to show the running job.

Once the job is finished, the job would set back to original status URL to show the test result.

Related: https://github.com/theopenlab/openlab/issues/311

### DIFF
--- a/zuul.d/pipelines.yaml
+++ b/zuul.d/pipelines.yaml
@@ -50,6 +50,7 @@
       github:
         status: pending
         comment: false
+        status-url: https://status.openlabtesting.org/status
     success:
       github:
         status: 'success'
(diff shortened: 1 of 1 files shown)


zuul.d/pipelines.yaml
```diff
@@ -50,6 +50,7 @@
       github:
         status: pending
         comment: false
+        status-url: https://status.openlabtesting.org/status
```
wangxiyuan: all pipelines should add this as well IMO.

Yikun: done

PR #585, opened 2019-07-10
Don't enable octavia on unsupported branches

huangtianhua:
Service octavia can not be installed successful on
stable/mitaka, stable/newton and stable/queens of
OpenStack, so don't enable it on these branches.

Related-Bug: theopenlab/openlab#289


roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -211,7 +203,8 @@
       EOF
     executable: /bin/bash
   when:
-    - '"octavia" in enable_services'
+    ('"octavia" in enable_services') and
+    (global_env.OS_BRANCH != 'stable/mitaka' or global_env.OS_BRANCH != 'stable/newton' or global_env.OS_BRANCH != 'stable/queens')
```
bzhaoopenstack: I'm afraid this condition won't filter any branches, if OS_BRANCH is stable/mitaka, then the first will false, but the second will true, and the whole condition will True.

kiwik: I think @bzhaoopenstack is right, could we use `not in`?

```python
global_env.OS_BRANCH not in ['stable/mitaka','stable/newton','stable/queens']
```

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -190,16 +190,8 @@
...
-          services='octavia,o-cw,o-hm,o-hk,o-api,q-lbaasv2'
-          cat << EOF >> /tmp/dg-local.conf
-          disable_service q-lbaas
-          enable_plugin neutron-lbaas https://opendev.org/openstack/neutron-lbaas
-      EOF
-      else
-          services='octavia,o-cw,o-hm,o-hk,o-api'
-      fi
+
+      services='octavia,o-cw,o-hm,o-hk,o-api'
```
kiwik: Please add echo message to show octavia is not available for some OpenStack releases in testing log, somebody maybe feel strange why they enable `octavia` in playbook, but in actually octavia can't be install.

kiwik: I don't see any updates about output message I mention in my 3 days ago comment

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -211,6 +203,9 @@
       EOF
     executable: /bin/bash
   when:
+    - global_env.OS_BRANCH != 'stable/mitaka'
```
kiwik: @liusheng Please help to review the PR, thanks

huangtianhua: I modified the name of the task, if don't match the conditions, the code won't execute, I think just the name of the task print, so I modified the task name. Is it ok?

roles/create-devstack-local-conf/tasks/main.yml
liusheng: if this task skiped by condition, it will not actually run, so `echo` message cannot work. maybe following is better:
`- name: create devstack local conf with octavia enabled for {{ global_env.OS_BRANCH }} release(M, N, Q unsupported)`

roles/create-devstack-local-conf/tasks/main.yml
liusheng: yes, see example:
```
2019-07-21 20:02:46.168005 | TASK [create-devstack-local-conf : create devstack local conf with Heat enabled]
2019-07-21 20:02:46.306743 | ubuntu-trusty | skipping: Conditional result was False
```

roles/create-devstack-local-conf/tasks/main.yml
liusheng: yes, `not in` statement looks better

roles/create-devstack-local-conf/tasks/main.yml
kiwik: OK, thank you.

roles/create-devstack-local-conf/tasks/main.yml
kiwik: OK, thank you.

PR #587, opened 2019-07-15
Do not use `no_log` and `register` together

wangxiyuan:
If using `no_log` and `register` together, the `regitster` one will be empty.


liusheng:
lgtm

PR #589, opened 2019-07-15
Apply consistent arch name "arm64"

kiwik:
1. Change job metadata from aarch64 to arm64
2. Remove useless nodeset "ubuntu-bionic-vxh-arm64"

Related-Bug: theopenlab/openlab#318


zuul.d/jobs.yaml
```diff
@@ -2173,7 +2173,7 @@
       - Project:apache/hbase
       - Application:Hbase@master
       - OS:ubuntu-xenial
-      - Arch:aarch64
+      - Arch:arm64
       - BuildType:Compile
```
bzhaoopenstack: looks duplicated with https://github.com/theopenlab/openlab-zuul-jobs/pull/588

zuul.d/jobs.yaml
```diff
@@ -2189,5 +2189,5 @@
       - Project:apache/hive
       - Application:Hive@master
       - OS:ubuntu-xenial
-      - Arch:aarch64
+      - Arch:arm64
```
bzhaoopenstack: And here, looks duplicated with https://github.com/theopenlab/openlab-zuul-jobs/pull/588

zuul.d/jobs.yaml
huangtianhua: Maybe we remain 'aarch64' for Arch metadata? We specify the 'Arch' comes from uname -a in redesign doc, or modify the doc to just tell we use arm64?

kiwik: @huangtianhua I have updated all of the related documents, include: redesign and docs.openlabtesting.org, please doube check it, thank you. And I think `Arch:arm64` is clear enough.

zuul.d/jobs.yaml
kiwik: Yep, I will rebase if #588 is merge first.

zuul.d/jobs.yaml
huangtianhua: OK, let's set it to arm64, and I will change related in my spark job.

PR #594, opened 2019-07-26
Do not enable plugin twice

huangtianhua:
Some openstack plugin enabled twice when install
devstack, refactor the code to avoid to enable
plugin twice.

Related-Bug: theopenlab/openlab#320


liusheng:
does this has been tested ?

PR #595, opened 2019-07-26
Add temp cpo jobs for test

wangxiyuan:
We want to move cpo related jobs from vexxhost to citynetwork. Before moving, we should test the jobs works well on citynetwork. This PR add temp jobs for testing. After testing well, we'll adopt the change to the offical jobs and remove the temp ones.

Related-Bug: theopenlab/openlab#321

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -2189,3 +2189,172 @@
       - OS:ubuntu-xenial
       - Arch:arm64
       - BuildType:Compile
+
+# Temp Kubernetes cloud-provider-openstack jobs on citynetwork cloud, these
+# jobs will be removed later.
+- job:
+    name: cloud-provider-openstack-test-citynetwork
+    parent: golang-test
+    description: |
+      Base job for all types of cloud-provider-openstack test jobs
+    pre-run: playbooks/cloud-provider-openstack-test/pre.yaml
+    post-run: playbooks/cloud-provider-openstack-test/post.yaml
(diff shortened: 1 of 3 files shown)


zuul.d/jobs.yaml
```diff
@@ -2189,3 +2189,172 @@
       - OS:ubuntu-xenial
       - Arch:arm64
       - BuildType:Compile
+
+# Temp Kubernetes cloud-provider-openstack jobs on citynetwork cloud, these
+# jobs will be removed later.
```
kiwik: Nice comments

zuul.d/jobs.yaml
```diff
@@ -2189,3 +2189,172 @@
...
+- job:
+    name: cloud-provider-openstack-test-citynetwork
+    parent: golang-test
+    description: |
+      Base job for all types of cloud-provider-openstack test jobs
+    pre-run: playbooks/cloud-provider-openstack-test/pre.yaml
+    post-run: playbooks/cloud-provider-openstack-test/post.yaml
+    nodeset: ubuntu-xenial-citynetwork
```
kiwik: Adding `nodeset` into abstract and non-running job looks not a good practice, because there is no `nodeset` define in child running job, so that cause developer is not easy to know which nodeset will be used, have to check parent job again, so I think we should add `nodeset` into running child directly, thought?

wangxiyuan: good idea. We can add them it to the offcial jobs in the next PR once the temp ones are well tested.

PR #603, opened 2019-08-01
Fix error of spark 2.4 k8s job

huangtianhua:
The job spark-integration-test-minikube-k8s is failed
due a reason "io.fabric8.kubernetes.client.KubernetesClientException:
404 page not found", and the spark community reply this
because too old a version of k8s we used, maybe some service
endpoints not existing.

This upgrades minikube to v0.34.1, upgrades kubernetes to v1.13.3.
And seems the tests passed, see the test patch:
https://github.com/theopenlab/spark/pull/14

Related-Bug: theopenlab/openlab#310


playbooks/spark-integration-test-minikube-k8s/run.yaml
```diff
@@ -3,7 +3,6 @@
   roles:
     - install-openjdk
     - role: install-docker
-      docker_version: 18.06
```
wangxiyuan: 19.03?

huangtianhua: I think latest is ok, now the latest release is 19.03, I don't know whether we have to hard it to 19.03.

wangxiyuan: If so,  the job tags should be `Application:Docker@latest` as well, otherwise once docker release a new version, the tag will be wrong.

huangtianhua: OK, I think set the docker version is more clear, but now there are many jobs have the problem, maybe we should fix them.

ZhengZhenyu:
LGTM

PR #610, opened 2019-08-09
Add scala install step

wangxiyuan:
Now many project use scala, it's good to add a common role.Since scala depends on java, add the install step inner openjdk role. Users can install scala after openjdk with the var `with_scala`

### DIFF
--- a/roles/install-openjdk/tasks/main.yaml
+++ b/roles/install-openjdk/tasks/main.yaml
@@ -30,3 +30,39 @@
   args:
     executable: /bin/bash
   environment: '{{ global_env }}'
+
+- name: Install scala
+  shell: |
+    set -xe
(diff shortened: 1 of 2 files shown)


roles/install-openjdk/tasks/main.yaml
```diff
@@ -30,3 +30,39 @@
...
+    tar -zxvf scala-{{ scala_version }}.tgz -C /usr/local
+  args:
+    executable: /bin/bash
+  when:
+    - with_scala|default(false)|bool
+
+- name: Set scala env vars
+  set_fact:
+    scala_env:
+      PATH: '/usr/local/scala-{{ scala_version }}/bin:{{ ansible_env.PATH }}'
```
Yikun: Do we also need a `SCALA_HOME` in here?

wangxiyuan: No, scala don't need it.

PR #616, opened 2019-08-12
Fix CPO job error

wangxiyuan:
upload_e2e.py uses subprocess.check_out(...encoding...). The encoding parameter is added in python3.6, but the default version in ubuntu16.04 is python3.5.

Install and use python3.7 to fix the issue.


ZhengZhenyu:
LGTM find same issue in kind job and fixed

adisky:
@wangxiyuan Thanks for the fix :)

PR #619, opened 2019-08-14
Fix CPO job openrc bug

wangxiyuan:
the `secrets` in the job definition is not inheritable. Create a new base job for lb related job on vexxhost provider.

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -708,18 +708,33 @@
     post-run: playbooks/cloud-provider-openstack-acceptance-test-manila-provisioner/post.yaml
     nodeset: ubuntu-xenial
 
+# TODO(wxy): Delete this base job once lb job is moved to citynetwork.
+- job:
+    name: cloud-provider-openstack-test-lb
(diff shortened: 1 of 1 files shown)


zuul.d/jobs.yaml
```diff
@@ -696,8 +691,6 @@
     run: playbooks/cloud-provider-openstack-acceptance-test-csi-cinder/run.yaml
     post-run: playbooks/cloud-provider-openstack-acceptance-test-csi-cinder/post.yaml
     nodeset: ubuntu-xenial-citynetwork
-    secrets:
-      - citynetwork_credentials
```
liusheng: if secrets cannot be inherited, will this also failed ?
https://github.com/theopenlab/openlab-zuul-jobs/blob/master/playbooks/cloud-provider-openstack-acceptance-test-csi-cinder/run.yaml#L4

did you run with this patch before merging?

wangxiyuan: oh,my fault.

ramineni:
can we get this merged asap to fix CI failures

PR #620, opened 2019-08-15
Fix ansible openstacksdk mitaka periodic job

bzhaoopenstack:
Openstack requirements already update the requests version to 2.22.0,
but pypi only has this specific version for python27 python35 python36
and python37. And currently we use py34 in our jobs. So we need to
upgrade the python version from py34 to py35.

test success in allinone:
https://github.com/bzhaoopenstack/ansible/pull/1

### DIFF
--- a/roles/install-devstack/tasks/main.yml
+++ b/roles/install-devstack/tasks/main.yml
@@ -33,7 +33,7 @@
       add-apt-repository ppa:deadsnakes/ppa -y
       apt-get update
       apt-get install python3.5 python3.5-dev libssl-dev -y
-      sudo update-alternatives --install $(which python3) python3 $(which python3.5) 1
+      update-alternatives --install $(which python3) python3 $(which python3.5) 1
 
       # Because the pip10 cannot uninstall/upgrade the pre-installed distutils packages
       # see: https://github.com/pypa/pip/issues/4805
--- a/roles/install-devstack/tasks/main.yml
+++ b/roles/install-devstack/tasks/main.yml
@@ -27,6 +27,14 @@
     cmd: |
       set -e
       set -x
+      # As openstack/requirements already upgrade requests to 2.22.0,
+      # but didn't provider a 2.22.0 version for python34 which we used.
+      # Install python3.5
+      add-apt-repository ppa:deadsnakes/ppa -y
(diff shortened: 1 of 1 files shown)


roles/install-devstack/tasks/main.yml
```diff
@@ -27,6 +27,14 @@
...
       set -e
       set -x
+      # As openstack/requirements already upgrade requests to 2.22.0,
+      # but didn't provider a 2.22.0 version for python34 which we used.
+      # Install python3.5
+      add-apt-repository ppa:deadsnakes/ppa -y
+      apt-get update
+      apt-get install python3.5 python3.5-dev libssl-dev -y
+      sudo update-alternatives --install $(which python3) python3 $(which python3.5) 1
+
```
wangxiyuan: I find there are at least 3 places installing python3.5 3.6 or 3.7 in the repo. We can add a new common  role for it.

roles/install-devstack/tasks/main.yml
```diff
@@ -27,6 +27,14 @@
     cmd: |
       set -e
       set -x
+      # As openstack/requirements already upgrade requests to 2.22.0,
+      # but didn't provider a 2.22.0 version for python34 which we used.
+      # Install python3.5
+      add-apt-repository ppa:deadsnakes/ppa -y
+      apt-get update
```
Yikun: do we also need the sudo in here? I didn't see any `become: yes` in here

roles/install-devstack/tasks/main.yml
bzhaoopenstack: Yeah, you, kevin and me already hit the issues.

roles/install-devstack/tasks/main.yml
bzhaoopenstack: Yeah, if we use sudo here may make this code stronger. But this is a role, the sudo priority should be setted in a job, maybe from a playbook. So here, we need to use become:yes in there.

bzhaoopenstack: To be honest, use sudo would be good for our code, but see other code in this file, they didn't use sudo, such as iptables need sudo. So that must be another PR to do refact.

PR #621, opened 2019-08-15
Add install python role

Yikun:
Related: https://github.com/theopenlab/openlab-zuul-jobs/pull/618
Related: https://github.com/theopenlab/openlab-zuul-jobs/pull/620


roles/config-python/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+python_major_version: '3'
```
wangxiyuan: this is unnecessary.
try:
```
python_version= {{ python_version }}
echo ${python_version:0:1}
```

Yikun: done. `major_version=`echo {{ python_version }} | cut -d '.' -f1``

roles/config-python/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+# We can install specify version gcc/g++ in task, like this:
```
liusheng: it is better to check if python3 has already installed and skip 'Install python' task if yes.

roles/config-python3/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+    set -ex
+    # 3.5 ==> 3
+    major_version=`echo {{ python_version }} | cut -d '.' -f1`
+    # Install python{{ python_version }}
+    add-apt-repository ppa:deadsnakes/ppa -y
+    apt-get update
+    apt-get install python{{ python_version }} python$major_version-dev libssl-dev -y
+    update-alternatives --install $(which python$major_version) python$major_version $(which python{{ python_version }}) 1
+  when:
+    - curr_python_version.stdout != python_version
```
liusheng: not sure if  version number `lower` better than `!=`

Yikun: As F2F discussion, It's better to match the specific version.

PR #624, opened 2019-08-24
Add csi-manila job

gman0:
This PR adds a playbook and a job definition for the CSI Manila acceptance test

### DIFF
--- a/playbooks/cloud-provider-openstack-acceptance-test-csi-manila/run.yaml
+++ b/playbooks/cloud-provider-openstack-acceptance-test-csi-manila/run.yaml
@@ -1,0 +1,201 @@
+- hosts: all
+  roles:
+    - config-golang
+    - export-cloud-openrc
+    - install-k8s
+  become: yes
+  tasks:
+    - name: Run csi manila acceptance tests with cloud-provider-openstack
+      shell:
+        cmd: |
+          set -x
+          set -e
+          set -o pipefail
+
+          #
+          # Preparation
+          #
+
+          export API_HOST_IP=$(ip route get 1.1.1.1 | awk '{print $7}')
+          export KUBELET_HOST="0.0.0.0"
+          export ALLOW_SECURITY_CONTEXT=true
+          export ENABLE_CRI=false
+          export ENABLE_HOSTPATH_PROVISIONER=true
+          export ENABLE_SINGLE_CA_SIGNER=true
+          export KUBE_ENABLE_CLUSTER_DNS=false
+          export LOG_LEVEL=4
+          export ALLOW_PRIVILEGED=true
+          # Just kick off all the processes and drop down to the command line
+          export ENABLE_DAEMON=true
+
+          # We need this for snapshots support, at least for now
+          export FEATURE_GATES="VolumeSnapshotDataSource=true"
+
+          mkdir -p '{{ k8s_log_dir }}'
+          export LOG_DIR='{{ k8s_log_dir }}'
+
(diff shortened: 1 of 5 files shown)


playbooks/cloud-provider-openstack-acceptance-test-csi-manila/run.yaml
```diff
@@ -0,0 +1,201 @@
+- hosts: all
+  roles:
+    - config-golang
+    - export-cloud-openrc
+    - install-k8s
+  become: yes
```
gouthampacha: Any reason we want to test this with a real cloud provider rather than with DevStack? Setup/run time?

The pre-setup you added to the current manila-provisioner acceptance test job should work here in case we could use DevStack:

https://github.com/theopenlab/openlab-zuul-jobs/blob/f12baea228cdbcc4186354d0e3e2f81b06c778bd/playbooks/cloud-provider-openstack-acceptance-test-manila-provisioner/run.yaml#L6-L14

gman0: Just quicker test runs

@gouthampacha and I just discussed this offline. The Citynetwork cloud doesn't seem to offer Manila anyway so we're going for a DevStack deployment

gman0:
PTAL @gouthampacha

roles/create-devstack-local-conf/tasks/main.yml
```diff
@@ -120,33 +120,25 @@
       set -e
       set -x
       cat << EOF >> /tmp/dg-local.conf
-      MANILA_MULTI_BACKEND=True
-      MANILA_BACKEND1_CONFIG_GROUP_NAME=london
-      MANILA_BACKEND2_CONFIG_GROUP_NAME=paris
-      MANILA_SHARE_BACKEND1_NAME=LONDON
-      MANILA_SHARE_BACKEND2_NAME=PARIS
+      # Enabling manila
+      enable_plugin manila https://git.openstack.org/openstack/manila
```
gouthampacha: Should work, but, "https://opendev.org/openstack/manila" would be preferable

zuul.d/jobs.yaml
```diff
@@ -708,6 +708,17 @@
...
+- job:
+    name: cloud-provider-openstack-acceptance-test-csi-manila
+    parent: cloud-provider-openstack-test
+    description: |
+      Run manila csi acceptance tests of cloud-provider-openstack
+    run: playbooks/cloud-provider-openstack-acceptance-test-csi-manila/run.yaml
+    post-run: playbooks/cloud-provider-openstack-acceptance-test-csi-manila/post.yaml
+    nodeset: ubuntu-xenial-citynetwork
+    secrets:
+      - citynetwork_credentials
```
gouthampacha: Any openlab host could work for this devstack based single node test job..

gman0:
@gouthampacha thanks for your feedback! I've removed the `nodeset` setting and changed the git repo url for Manila. PTAL

playbooks/cloud-provider-openstack-acceptance-test-csi-manila/run.yaml
```diff
@@ -0,0 +1,208 @@
+- hosts: all
+  roles:
+    - config-golang
+    - export-cloud-openrc
```
wangxiyuan: no need for this role if you don't use cloud service.

gman0: ack

PR #628, opened 2019-08-29
Add a commnet only pipeline

wangxiyuan:
Flink doesn't want ARM CI jobs run on every PR before it's stable enough. Add a comment only pipeline for this kind of requirement.

### DIFF
--- a/zuul.d/pipelines.yaml
+++ b/zuul.d/pipelines.yaml
@@ -1055,3 +1055,29 @@
       github:
         status: 'failure'
       mysql:
+
+- pipeline:
+    name: arm_comment_only
+    description: |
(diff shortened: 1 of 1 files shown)


zuul.d/pipelines.yaml
```diff
@@ -1055,3 +1055,29 @@
       github:
         status: 'failure'
       mysql:
+
+- pipeline:
+    name: arm_comment_only
+    description: |
+      This pipeline is triggered by github comment only for ARM test. Before CI
```
bzhaoopenstack: You'd better to tell us commenting What to trigger this pipeline. Or keep the same style with other pipelines in description.

PR #634, opened 2019-09-04
fix cpo jobs

adisky:
go get kubetest in not installing succesfully from
 $GOPATH/k8s.io/src/cloud-provider-openstack
need to run it from non repo

### DIFF
--- a/playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance/run.yaml
+++ b/playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance/run.yaml
@@ -103,6 +103,7 @@
           export KUBECONFIG=/var/run/kubernetes/admin.kubeconfig
 
           # Build kubetest
+          cd ..
           export GO111MODULE=on
           export GOPROXY=https://proxy.golang.org
           export GOSUMDB=sum.golang.org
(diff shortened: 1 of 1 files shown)


playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance/run.yaml
```diff
@@ -116,7 +117,7 @@
 
           # Run e2e test using local deployment/provider
           # Must run kubetest under kubernetes root directory
-          cd '{{ k8s_src_dir }}'
+          pushd '{{ k8s_src_dir }}'
```
ramineni: no need to change here I suppose

playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance/run.yaml
```diff
@@ -103,6 +103,7 @@
           export KUBECONFIG=/var/run/kubernetes/admin.kubeconfig
 
           # Build kubetest
+          cd ..
```
ramineni: May be add comment for future reference, your call ?

wangxiyuan: it's good to add a comment that why `cd` here.

playbooks/cloud-provider-openstack-acceptance-test-e2e-conformance/run.yaml
wangxiyuan: single using `pushd` is useless. just `cd` here is OK.

PR #642, opened 2019-09-10
Split the periodic pipeline to a single period

bzhaoopenstack:
For saving arm resources, we only run 1 time per day for ARM periodic testing.


bzhaoopenstack:
Spark long time job change to UTC-12, if it lasts 14 hours(max), it will be complete before we come in local time.

bzhaoopenstack:
Any thought about other ARM periodic jobs?

PR #648, opened 2019-09-16
[WIP]Add CPO CSI migration test

wangxiyuan:


adisky:
@wangxiyuan we need to enable feature gates here https://github.com/theopenlab/openlab-zuul-jobs/blob/master/roles/create-multinodes-k8s-cluster-with-kubeadm/tasks/main.yaml#L43
`--feature-gates=CSIMigration=true,CSIMigrationOpenStack=true`

wangxiyuan:
see https://github.com/theopenlab/openlab-zuul-jobs/pull/649

PR #651, opened 2019-09-19
Remove hyperkube reference

adisky:
Fix CI job for CPO, hyperkube removed from ./hack/local-up-cluster.sh


adisky:
@wangxiyuan could we merge this??

PR #655, opened 2019-09-20
Build hyperkube binary

ramineni:
hyperkube binary is removed from local-up-cluster.sh but still
used by at other places (golang.sh), so adding the same.


adisky:
@ramineni Thanks
/lgtm

PR #665, opened 2019-09-25
Update post job for kind to use periodic-6

ZhengZhenyu:
Kind job has been changed to periodic-6, this script here should also be updated.


liusheng:
looks good, but better to make the PR name more specific

PR #666, opened 2019-09-25
[WIP]Change apache/spark arm job to periodic-0

huangtianhua:
The spark arm job fails always these days, seems
the performance of arm instance choosen is poor.
This changes the periodic time of spark job to utc0.

### DIFF
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -44,8 +44,6 @@
             branches: master
         - spark-integration-test-minikube-k8s-v1.13.3:
             branches: master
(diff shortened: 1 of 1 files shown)


zuul.d/projects.yaml
```diff
@@ -261,3 +259,10 @@
             branches: master
         - flink-end-to-end-test-part3:
             branches: master
+
+- project:
```
bzhaoopenstack: I think it's better to move this block to the related Periodic. periodic-0 task should be under periodic-0/12 "####################### periodic jobs on 00:00/12:00 ##########################"

PR #676, opened 2019-10-08
Update cpo e2e conformance jobs to run with latest release.

ramineni:
This commit adds cpo e2e conformance job for 1.16 release and also removes
older releases than 1.14

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -683,6 +683,28 @@
       - BuildType:Integration test
 
 - job:
+    name: cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.16
+    parent: cloud-provider-openstack-acceptance-test-e2e-conformance
+    description: |
+      Run Kubernetes E2E Conformance tests against release-1.16 branch of Kubernetes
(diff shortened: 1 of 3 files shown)


lingxiankong:
/lgtm

thanks!

adisky:
thanks
/lgtm

zuul.d/jobs.yaml
```diff
@@ -682,6 +682,28 @@
       - Arch:x86_64
       - BuildType:Integration test
 
+- job:
```
huangtianhua: So if we don't need to test with k8s v1.13 any more? Maybe we should delete the job 1.13 definition above in jobs.yaml?

wangxiyuan: It's not easy to remove them since many CPO branches  contains them in .zuul.yaml. We should remove them from CPO side first.

ramineni: Actually, I tried that , it didnt allow me because this job is being used for other branches eg: on release1.15 / release1.14 still this job runs.

huangtianhua: OK, let's leave it as it is now.

PR #678, opened 2019-10-10
Fix conformance jobs

ramineni:
This PR fixes the conformance job failure, which is caused by unable to build kubetest


adisky:
thanks
/lgtm

PR #679, opened 2019-10-11
Bump go version to 1.12.10

kayrus:
This PR fixes the conformance job failure, which is caused by unable to build kubernetes:

```
make: Entering directory '/home/zuul/src/k8s.io/kubernetes'
make[1]: Entering directory '/home/zuul/src/k8s.io/kubernetes'

Detected go version: go version go1.12.1 linux/amd64.
Kubernetes requires go1.12.4 or greater.
Please install go1.12.4 or later.

!!! [1011 04:43:22] Call tree:
!!! [1011 04:43:22]  1: hack/run-in-gopath.sh:30 kube::golang::setup_env(...) 

Detected go version: go version go1.12.1 linux/amd64.
Kubernetes requires go1.12.4 or greater.
Please install go1.12.4 or later.
```

/cc @adisky @ramineni @huangtianhua

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -560,6 +560,8 @@
     run: playbooks/cloud-provider-openstack-acceptance-test-k8s-cinder/run.yaml
     post-run: playbooks/cloud-provider-openstack-acceptance-test-k8s-cinder/post.yaml
     nodeset: ubuntu-xenial-citynetwork
+    vars:
+      go_version: '1.12.10'
     secrets:
       - citynetwork_credentials
 
@@ -699,6 +701,8 @@
     run: playbooks/cloud-provider-openstack-acceptance-test-csi-cinder/run.yaml
     post-run: playbooks/cloud-provider-openstack-acceptance-test-csi-cinder/post.yaml
     nodeset: ubuntu-xenial-citynetwork
+    vars:
+      go_version: '1.12.10'
     secrets:
       - citynetwork_credentials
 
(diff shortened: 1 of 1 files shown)


zuul.d/jobs.yaml
```diff
@@ -645,7 +645,7 @@
       Run Kubernetes E2E Conformance tests against release-1.14 branch of Kubernetes
     nodeset: ubuntu-xenial-citynetwork
     vars:
-      go_version: '1.12.1'
```
ramineni: This doesnt need to change , release-1.14 supports 1.12.1

zuul.d/jobs.yaml
```diff
@@ -667,7 +667,7 @@
       Run Kubernetes E2E Conformance tests against release-1.15 branch of Kubernetes
     nodeset: ubuntu-xenial-citynetwork
     vars:
-      go_version: '1.12.1'
```
ramineni: ditto

zuul.d/jobs.yaml
kayrus: It's a minor version, besides 1.12.10 contains security fixes: https://golang.org/doc/devel/release.html#go1.12

kayrus:
fixed

adisky:
@huangtianhua @wangxiyuan @ramineni @kayrus  kubernetes updated minimum go version to 1.12.4, i think we also should update the default version
https://github.com/theopenlab/openlab-zuul-jobs/blob/master/roles/config-golang/defaults/main.yml

if openlab team is fine with updating default version, we dont need to pin version in these tests.
@huangtianhua @wangxiyuan please suggest

adisky:
as discussed on irc, lets merge this, default version is upgraded after consensus from all projects.
/lgtm

PR #682, opened 2019-10-15
WIP: update KUBELET_HOST to contain actual ip

ramineni:
Recent changes in k8s , https://github.com/kubernetes/kubernetes/pull/82413
checks for KUBELET_HOST in get nodes info, which is resulting in
error. This commit is to update the same.


ramineni:
cc @adisky

PR #686, opened 2019-10-23
Refactor flink cron job

wangxiyuan:
1. Update e2e test name
2. Add unit test job
3. Add a common playbook.


playbooks/flink-end-to-end-test/build_and_test_core.yaml
```diff
@@ -0,0 +1,54 @@
+- import_playbook: common.yaml
+
+# NOTE: Some tests should not be ran under root.
+- hosts: all
+  become: no
```
liusheng: if non root users, the following `apt ...` command should be added `sudo` prefix ?

PR #704, opened 2019-11-07
CPO: Update csi tests

ramineni:
This commit is to install kubectl binary explictly as needed by
e2e test and also updates the expression for pod status checking.


ramineni:
cc @adisky @wangxiyuan

PR #705, opened 2019-11-07
Remove deprecated hyperkube from the make target

kayrus:
Hyperkube has been deprecated, see https://github.com/kubernetes/kubernetes/pull/84662
/cc @adisky @ramineni

### DIFF
--- a/roles/install-k8s/tasks/main.yaml
+++ b/roles/install-k8s/tasks/main.yaml
@@ -17,6 +17,6 @@
 
       # Build k8s cmd
       git clone https://github.com/kubernetes/kubernetes '{{ k8s_src_dir }}' -b '{{ k8s_version }}'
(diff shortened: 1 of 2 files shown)


roles/install-k8s/tasks/main.yaml
```diff
@@ -17,6 +17,6 @@
 
       # Build k8s cmd
       git clone https://github.com/kubernetes/kubernetes '{{ k8s_src_dir }}' -b '{{ k8s_version }}'
-      make -C '{{ k8s_src_dir }}' WHAT="cmd/kubectl cmd/hyperkube cmd/kube-apiserver cmd/kube-controller-manager cmd/cloud-controller-manager cmd/kubelet cmd/kube-proxy cmd/kube-scheduler"
+      make -C '{{ k8s_src_dir }}' WHAT="cmd/kubectl {% if not (k8s_version in ['master', 'release-1.17']) %} cmd/hyperkube{% endif %} cmd/kube-apiserver cmd/kube-controller-manager cmd/cloud-controller-manager cmd/kubelet cmd/kube-proxy cmd/kube-scheduler"
```
riking: won't this break in release-1.18?

kayrus: updated

PR #649, opened 2019-09-18
[WIP] Add CPO migration job

adisky:

PR #707, opened 2019-11-09
Implemented k8s e2e.test for cinder CSIMigration

chrigl:
This job runs the sig-storage + driver: cinder e2e test suite against a
multinode Kubernetes cluster. The cluster is configured with enabled
feature-gates CSIMigration, CSIMigrationOpenStack and ExpandCSIVolumes
on kube-apiserver, kube-controller-manager and kubelet.

This is the replacement for #649
 
Close-bug: theopenlab/openlab#342


wangxiyuan:
Thanks, Just want to make sure that this job will be ran as periodic job, right?

chrigl:
Thanks @wangxiyuan. This job runs up to an hour, so I think in the first step it makes sense to have that periodically. Maybe I will be able to extract some really essential tests and run them on a per PR basis.

PR #721, opened 2019-11-15
CPO: Fix keystone job error

ramineni:
CPO keystone job is failing 
Root cause: kube-apiserver failed to start with error Error: invalid authorization config: unsupported webhook authorizer version "", supported versions are "v1", "v1beta1" . 
Explicitly pass the version while starting apiserver as per https://github.com/kubernetes/kubernetes/pull/84768


ramineni:
cc @adisky

adisky:
lgtm

PR #740, opened 2019-12-09
Switch to use nodeset ubuntu-bionic for terraform jobs

huangtianhua:
Related-Bug: theopenlab/openlab#412


dtantsur:
Hi, has this fixed your problem? I'm doing seemingly the same and getting NODE_FAILURE: https://github.com/dtantsur/rust-openstack/pull/108/

PR #755, opened 2019-12-23
csi-manila job: fixed kubectl exec, escaped eval

gman0:
Fixed two issues that didn't show up during local tests:
* used hard-coded path for kubectl
* `$(NODE_ID)` in heredoc needs to be escaped


gman0:
PTAL @gouthampacha @wangxiyuan

PR #774, opened 2020-01-17
CPO: update the pipeline for CPO jobs

ramineni:
Pipeline for CPO jobs seems to be changed, so latest images are
not getting pushed to docker registry. This commit updates the
pipeline in CPO jobs as well to fix the same.
This commit also fixes the periodic job failure of 
cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.16
by updating go version to 1.13.5 .


ramineni:
cc @lingxiankong @chrigl

wangxiyuan:
BTW, is it enough to run the periodic  once a day? if not, i can get it back to twice

ramineni:
> BTW, is it enough to run the periodic once a day? if not, i can get it back to twice

Once should be fine. Thanks.

PR #781, opened 2020-01-24
Set insecure registry to docker configuration

ramineni:
This PR enables insecure registry to get registered with docker configuration


wangxiyuan:
@ramineni  Sorry that I am Holiday. Can't reply at once. Just ping me if you hit problem

PR #786, opened 2020-01-30
install-devstack: Removed python3-httplib2 with apt module

chrigl:
Currently apt waits for user input to remove `python3-httplib2`. This PR uses the apt module to uninstall the package.

Relates to: #785 

https://logs.openlabtesting.org/logs/19/919/8c9146151bd70313453c38328cc51dd64af61c0c/cloud-provider-openstack-acceptance-test-flexvolume-cinder/cloud-provider-openstack-acceptance-test-flexvolume-cinder/aae2d4a/job-output.txt.gz

```
2020-01-30 04:16:14.732684 | ubuntu-bionic | The following packages will be REMOVED:
2020-01-30 04:16:14.734138 | ubuntu-bionic |   apport python3-apport python3-httplib2 ubuntu-server
2020-01-30 04:16:14.742028 | ubuntu-bionic | 0 upgraded, 0 newly installed, 4 to remove and 3 not upgraded.
2020-01-30 04:16:14.742122 | ubuntu-bionic | After this operation, 1533 kB disk space will be freed.
2020-01-30 04:16:14.742209 | ubuntu-bionic | Do you want to continue? [Y/n] Abort.```

PR #787, opened 2020-01-30
Fix the fix for #452

tombarron:
I should have used '-y' with 'apt remove' since this is a non-interactive context.


tombarron:
https://github.com/theopenlab/openlab-zuul-jobs/pull/786 is a better approach to this.

PR #790, opened 2020-02-04
CPO: included export-cloud-openrc

chrigl:
Credentials were not set in tests, because the role `export-cloud-openrc` was missing.


ramineni:
LGTM

PR #808, opened 2020-02-28
Use ``demo`` user/project for csi-manila-secrets

tombarron:
Using the ``admin`` user and project allows false-positive test results if the code being tested uses OpenStack APIs that are not allowed for non-administrators, so use the regular ``demo`` user and project set up by devstack instead.

### DIFF
--- a/playbooks/cloud-provider-openstack-acceptance-test-csi-manila/run.yaml
+++ b/playbooks/cloud-provider-openstack-acceptance-test-csi-manila/run.yaml
@@ -140,12 +140,16 @@
           metadata:
             name: csi-manila-secrets
             namespace: default
+          # Use the devstack ``demo`` user in the ``demo`` project rather than ``admin``
+          # since regular users should be able to use the cloud-provider-openstack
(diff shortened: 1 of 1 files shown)


tombarron:
@gman0 ptal

gman0:
Thanks

/lgtm

gouthampacha:
@lingxiankong - can you please review this

playbooks/cloud-provider-openstack-acceptance-test-csi-manila/run.yaml
```diff
@@ -140,12 +140,16 @@
           metadata:
             name: csi-manila-secrets
             namespace: default
+          # Use the devstack ``demo`` user in the ``demo`` project rather than ``admin``
+          # since regular users should be able to use the cloud-provider-openstack
+          # APIs without OpenStack administrative privileges.  Devstack sets up
+          # up the same password for both users.
```
wangxiyuan: It's good to make the user configurable in `install-devstack` role. While the role doesn't support it currently, I'm OK with this change.

PR #811, opened 2020-03-12
Fix CPO resource cleanup bug

wangxiyuan:


playbooks/cloud-provider-openstack-test/post.yaml
```diff
@@ -16,11 +16,14 @@
           fi
           pv_name=$('{{ kubectl }}' get pv | sed '1d' | awk '{print $1}')
           if [[ -n "$pv_name" ]]; then
-            volume_id=$(openstack volume list | awk "/$pv_name/ {print \$2}")
+            for i in $pv_name
+            do
+            volume_id=$(openstack volume list | awk "/$i/ {print \$2}")
             if [[ -n "$volume_id" ]]; then
                 openstack server remove volume "$server_id" "$volume_id"
```
lingxiankong: Do we need to wait for the volume available before removal?

wangxiyuan: yeah, let me add the check later.

PR #812, opened 2020-03-12
Fix cloud-provider-openstack-e2e-test-csi-cinder job

ricolin:
Due to [1], we need to update this job to use new way to build images.

[1] https://github.com/kubernetes/cloud-provider-openstack/pull/910


ramineni:
thanks . +1

PR #834, opened 2020-03-25
[CPO][fix job] avoid using docker manifest with local repository

ricolin:
Avoid using docker manifest with local image repository


ramineni:
Thanks. lgtm

PR #854, opened 2020-04-07
OCCM: deploy occm for topology testing in v1.18.0

chrigl:
From Kubernetes v1.18.0 on, the topology support and e2e testing is
extended. Therefor this commit also deploys
openstack-cloud-controller-manager to get nodes labeled with openstack
availability zone and region.

I needed to move the generation of the cloud-config secret up, so that occm
has access during its deployment, so that nodes can be labeled
appropriately, and a NoSchedule taint is removed.

### DIFF
--- a/playbooks/cloud-provider-openstack-multinode-csi-migration-test/pre.yaml
+++ b/playbooks/cloud-provider-openstack-multinode-csi-migration-test/pre.yaml
@@ -1,6 +1,8 @@
 - hosts: k8s-master
   name: prepare configuration of kubeadm and set CSIMigration feature-gates
   become: yes
+  roles:
+    - export-cloud-openrc
   tasks:
     - name: create /etc/kubernetes
       file:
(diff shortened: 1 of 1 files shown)


playbooks/cloud-provider-openstack-multinode-csi-migration-test/pre.yaml
```diff
@@ -18,14 +20,11 @@
...
+              "cloud-provider": "external"
           controllerManager:
             extraArgs:
               "feature-gates": "CSIMigration=true,CSIMigrationOpenStack=true,ExpandCSIVolumes=true"
-          ---
-          apiVersion: kubeadm.k8s.io/v1beta2
-          kind: InitConfiguration
-          nodeRegistration:
-              name: {{ inventory_hostname }}
+              "cloud-provider": "external"
```
ramineni, line 27: @chrigl im confused, if we add cloud-provider as external, we wont be using intree anymore?

chrigl: In v1.18.0 we have new tests in the Kubernetes e2e suite. These ones now fail, because the test-cluster does not have proper topology labels assigned to the nodes. These could be either done by the old in-tree cloud-provider, or OCCM, or manually by hand. I decided to use OCCM, because the in-tree one is going away in the future, but more important, I don't want to pass in OS credentials into kube-controller-manager and kubelet. So we are really sure, the CSIMigration is used.

ramineni: got it. Thanks :)

ramineni:
@wangxiyuan please review

PR #957, opened 2020-08-05
Fix gpdb ci test env

bzhaoopenstack:
In test env, I found it will inject the interaction during demo cluster
startup. So for forbid this issue, add the localhost signal into
known_hosts when the cluster setup the segment using hostname.


playbooks/gpdb-installcheck-world-tests-on-arm64/run.yaml
```diff
@@ -161,9 +161,9 @@
         executable: /bin/bash
         chdir: /home/zuul/gpdb
 
-    - name: debug
+    - name: Genarate local known_hosts for demo cluster startup
```
martin-g, line 164: minor typo in gen`e`rate

PR #971, opened 2020-08-18
Fix go version to 1.15

ramineni:
1.15.0 version key is not found , needs to be updated to 1.15 to get downloaded.
This commit fix the same.


wangxiyuan:
Right

PR #976, opened 2020-08-19
Add mysql periodic test

bzhaoopenstack:
This proposal introduces the mysql timer build on UTC2:00 per Sunday.
This framework will work on the upstream version and optimized version,
but for now, the optimized one is not ready. So enable the upstream one
first.

For now, only enable mysql 5.7.27 upstream version.


bzhaoopenstack:
recheck

PR #992, opened 2020-08-24
Adding support for train and ussuri triggers

jtopjian:
For: https://github.com/theopenlab/openlab/issues/574

### DIFF
--- a/zuul.d/pipelines.yaml
+++ b/zuul.d/pipelines.yaml
@@ -238,6 +238,56 @@
         status: 'failure'
       mysql:
 
+- pipeline:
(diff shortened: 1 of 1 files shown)


bzhaoopenstack (approved):
LGTM, and this needs to restart the zuul server to load the new config. Please wait. Once we finish to restart, will reply on this PR.

wangxiyuan (approved):

wangxiyuan:
LGTM, then you can use these pipelines in https://github.com/gophercloud/gophercloud/blob/master/.zuul.yaml

PR #1091, opened 2021-01-29
Make install-k8s role for supporting customize docker version

bzhaoopenstack:


jichenjc:
https://github.com/kubernetes/cloud-provider-openstack/pull/1320

seems works fine now

PR #1104, opened 2021-02-06
Revert "Add missing no-log to install-devstack (#1062)"

jeblair:
This reverts commit 30ffe4b62a41037d08491aaccf7c39f2b1b2e1a7.

Terraform and gophercloud jobs are failing on this task[1], and the
no_log hides the actual error.  We don't really need a no_log
here since this openrc is for a throwaway test cloud -- the password
that is being hidden is the literal string "password" so there's
no secret info.

[1] https://github.com/theopenlab/openlab/issues/681

### DIFF
--- a/roles/install-devstack/tasks/main.yml
+++ b/roles/install-devstack/tasks/main.yml
@@ -111,7 +111,6 @@
       echo "{$openrc}"
     executable: /bin/bash
   register: output
-  no_log: yes
 
 - name: Set fact for devstack openrc
   set_fact:


emonty (approved):
There is no need to no_log this.

wangxiyuan (approved):

dtantsur:
> There is no need to no_log this.

Well, it outputs passwords in plain text.

emonty:
On Mon, Feb 8, 2021, 04:50 Dmitry Tantsur <notifications@github.com> wrote:

> There is no need to no_log this.
>
> Well, it outputs passwords in plain text.
>

Yeah, but they're not real passwords. They're devstack passwords. Is there
a context in which this is being used with actually consequential passwords?

In any case, amazingly the no_log actually breaks the jobs.

wangxiyuan:
yeah, I think this password is just a temp one for devstack env. The env be deleted once the job is done. And I think the devstack env is not exposed outside. So there is no security problem IMO

I approved the `no_log` patch previously because that I thought it's just a tiny change. I didn't realize that it will break some jobs.

PR #1113, opened 2021-02-16
[CPO] added new bucket

chrigl:
We must replace our old bucket for test results by a new one. In this
PR, I added the new one additionally to the old one. When the new one
works as expected, I will remove the old one.

See: https://github.com/kubernetes/k8s.io/issues/1311


chrigl:
Blocked by #1114

wangxiyuan (approved):

wangxiyuan:
ansible-lint problem should not block this PR.

PR #1114, opened 2021-02-16
Fixed installation of ansible-lint

chrigl:
ansible-lint released version 5.0.0 what doesn't depend hard on ansible
any more. This PR installs ansible-lint[core] that does depend on
ansible.
See
https://github.com/ansible-community/ansible-lint/blob/650bc07fc4cbd05f9a98ea614589f907e965c61d/setup.cfg#L96


chrigl:
I don't have any idea why the test-run doen't use the code of my change. Maybe @bzhaoopenstack could help out?

PR #1121, opened 2021-02-18
WIP recrypted gcp_account_cpo due to POST_FAILURE

chrigl:
Probably, before probably I used a wrong key. Unfortunately I am not able to debug it.

Now using
```
encrypt_secret.py --infile service-provider-openstack-k8s-key.json --tenant openlab https://status.openlabtesting.org kubernetes/cloud-provider-openstack
```

Before I used the same with `theopenlab/openlab-zuul-jobs`.

However, if this again is the wrong one, I would need some help :)
Thank you!

### DIFF
--- a/zuul.d/secrets.yaml
+++ b/zuul.d/secrets.yaml
@@ -562,56 +562,56 @@
     name: gcp_account_cpo
     data:
       key_json: !encrypted/pkcs1-oaep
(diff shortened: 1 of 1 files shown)


chrigl:
Seems not to be the encryption... opened another PR for getting debug info.

PR #1131, opened 2021-03-09
[CPO] Add signal file for PR label

lingxiankong:
When job A is running but the PR label was set by another job B, the job A should wait. However, if the waiting times out, job A shouldn't remove the label which was set by another job.

The Kubernetes resources should be cleaned up by the script itself in CPO repo.


lingxiankong:
@jichenjc @ramineni FYI

bzhaoopenstack (approved):

jichenjc:
@lingxiankong thanks for sharing~

PR #1137, opened 2021-03-25
Disable gophercloud

wangxiyuan:
Disable gophercloud

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -372,106 +372,106 @@
 
 ## ========== gophercloud + openstack jobs ==============
 ## grandson job for stein
-- job:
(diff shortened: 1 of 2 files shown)


wangxiyuan:
I assume gophercloud has removed openlab app. Just disable the related jobs currently. Let me know if openlab is still used by gophercloud.

jtopjian:
@wangxiyuan We had no intention to stop using OpenLab. I did just notice that Gophercloud pull requests are not triggering OpenLab - this is probably an error of some sort, but it appears that this just started happening in the last few days.

Can you please add the Gophercloud jobs back to OpenLab so I can investigate?

wangxiyuan:
@jtopjian OK, we'll reenable it soon

PR #1149, opened 2021-05-19
cleanup for orphaned volumes of cinder csi

chrigl:
Discussed in https://github.com/theopenlab/openlab/issues/746

This PR runs a job periodically to remove orphaned volumes created by
cinder csi. Post cleanup jobs directly on the real test jobs are not
robust enough in cases of e.g. timeouts or random crashes.

### DIFF
--- a/roles/cleanup-cinder-csi-volumes/files/cleanup_volumes.py
+++ b/roles/cleanup-cinder-csi-volumes/files/cleanup_volumes.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+"""
+This script removes volumes created by cinder csi that are
+* older than 3 hours
+* not attached to any node
+"""
+import openstack
+import logging
+import sys
+from datetime import datetime, timedelta
+
+def main():
+
+    logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+    log = logging.getLogger(__name__)
+    log.info("cleaning up orphaned volumes")
+
+    con = openstack.connect()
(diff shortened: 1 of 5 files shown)


wangxiyuan (approved):

wangxiyuan:
@bzhaoopenstack please take a look at this one. This need a zuul restart as well IMO

roles/cleanup-cinder-csi-volumes/files/cleanup_volumes.py
```diff
@@ -0,0 +1,52 @@
...
+
+    for v in filter(filter_volumes, volumes):
+        log.info("deleting volume %s", v.id)
+        try:
+            con.block_storage.delete_volume(v, ignore_missing=True)
+        except:
+            log.exception("unable to delete volume %s... skipping", v.id)
+            pass
+
+def filter_volumes(vol):
```
lingxiankong, line 29: @wangxiyuan @bzhaoopenstack please take a look at the function here to make sure it doesn't remove volumes that shouldn't be removed.

bzhaoopenstack: Thanks, I think this function is OK  to  go. As current volumes which is used in current env are not created with the name prefix "pvc-", and if the volume is using by other VM, those ones can not be deleted.

bzhaoopenstack (approved):
Thanks, lgtm

PR #1188, opened 2021-06-10
[CPO] Use cloud-provider-openstack-test-occm as base job

lingxiankong:
This change only affects cloud-provider-openstack-acceptance-test-e2e-conformance-stable-branch-v1.21

### DIFF
--- a/playbooks/cloud-provider-openstack-test-occm/run-1.19.yaml
+++ b/playbooks/cloud-provider-openstack-test-occm/run-1.19.yaml
@@ -48,7 +48,7 @@
         name: install-k3s-vm
       vars:
         devstack_workdir: "{{ global_env.devstack_workdir }}"
-        cpo_branch: "{{ zuul.branch }}"
+        k8s_branch: "{{ k8s_release | default(zuul.branch) }}"
         worker_node_count: 1
 
     - name: Deploy openstack-cloud-controller-manager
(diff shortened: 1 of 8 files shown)


playbooks/cloud-provider-openstack-test-occm/run-1.19.yaml
```diff
@@ -48,7 +48,7 @@
         name: install-k3s-vm
       vars:
         devstack_workdir: "{{ global_env.devstack_workdir }}"
-        cpo_branch: "{{ zuul.branch }}"
+        k8s_branch: "{{ k8s_release | default(zuul.branch) }}"
```
ramineni, line 51: @lingxiankong etcd versions /go lang versions supported change for each k8s release. Does this prerequisites taken care by k3s?

lingxiankong: What's the requirement for etcd and golang version? k3s doesn't use etcd. Why do we care about Golang? We have openstack-cloud-controller-manager built with minimum version of Go 1.15, and openstack-cloud-controller-manager talks with k8s API.

ramineni: Ah, right. We are not building k8s unlike before, so shouldnt be required.

ramineni:
looks good to me

lingxiankong:
> looks good to me

Thanks for review, I will merge and restart zuul.

PR #1206, opened 2021-11-02
Add a workaround to avoid issues when devstack needs APT

EmilienM:
It's rare but it can happen that devstack runs at the same time as a
crontab is triggered and runs APT to e.g. execute unattended upgrades of
the system.

Devstack would fail since APT isn't available.
This patch adds a workaround so it'll remove the crontabs before
deploying devstack.


EmilienM:
recheck

EmilienM:
@wangxiyuan when CI is green please have a look.

wangxiyuan (approved):

EmilienM:
sorry I had to rebase on master.

wangxiyuan (approved):

PR #1211, opened 2021-11-03
Use devstack_env into create-devstack-local-conf role

EmilienM:
If devstack_env is set, we'll add the environment variables into the
local config for devstack. This will be useful for anyone who wants to
configure devstack for their needs, in the CI jobs.

This is backward compatible and disabled by default.


EmilienM:
ping @wangxiyuan

wangxiyuan (approved):

PR #1212, opened 2021-11-04
create-devstack-local-conf: skip overrides if devstack_env is empty

EmilienM:
In my previous patch, I missed the fact that we set a default for
`devstack_env`. This patch will helps to make the playbook faster, if we
don't need to run the block. If the variable is an empty dic, we'll skip
the block as there is no need to configure overrides.


EmilienM:
ping @wangxiyuan - a follow-up from the last patch. Thanks

wangxiyuan (approved):

EmilienM:
recheck

PR #1214, opened 2021-11-05
Make sure we uninstall unattended-upgrades on Ubuntu

EmilienM:
unattended-upgrades package in Ubuntu sets up a crontab that will
periodically upgrade the system and therefore run "apt".

We don't want that because it'll lock apt so things like devstack will
fail since they won't be able to use apt.

Let's just remove the package if installed, which is how other infras
are doing.


EmilienM:
ping @wangxiyuan

wangxiyuan (approved):

PR #1215, opened 2021-11-06
Add a role to prefetch Amphora from RDO server

EmilienM:
Instead of building Amphora, allow to download it from a place (default
is RDO), into a specific directory.

This will save time for some jobs where Octavia devstack plugin builds
the Amphora image.


EmilienM:
ping @wangxiyuan

wangxiyuan (approved):

